PR #55, opened 2015-02-18
Fixed problem with breakpoints not being applied for custom rules

renatoi:
@thierryk @src-code


yahoocla:
CLA is valid!

thierryk:
Nope! That did not fix it :(

PR #56, opened 2015-02-18
Updated dev dependencies

renatoi:
Now that sinon-chai has been updated to support chai 2.0.0.


yahoocla:
CLA is valid!

src-code:
:shipit:

PR #59, opened 2015-02-20
Suffix to value

renatoi:
@src-code @thierryk


yahoocla:
CLA is valid!

src-code:
:shipit:

PR #63, opened 2015-02-28
Integrating AutoAtom functionality

src-code:
Bringing this over from grunt-atomizer... Note this is not BC, and will result in a major version bump

### DIFF
--- a/bin/atomizer
+++ b/bin/atomizer
@@ -14,14 +14,28 @@ var path = require('path');
 var fs = require('fs');
 var chalk = require('chalk');
 var atomizer = require('../src/atomizer');
-var configObjs = [];
+var _ = require('lodash');
+var content = '';
+var config; 
+var parsedConfig = {};
 
 var params = require('minimist')(process.argv.slice(2));
 
 if (process.argv.slice(2).length === 0 || params.help) {
-    var usage = ['usage:  ', process.title, '[-o|--outfile=<file>] [--help] configfile ...'].join(' ');
+    var usage = ['usage:  ', process.title, ' -c|--config=<file> [-o|--outfile=<file>] [--help] [--verbose] [<files-to-parse> ...]'].join(' ');
     console.log(usage);
     return;
+}
+
+function handleMergeArrays (a, b) {
+    if (_.isArray(a) && _.isArray(b)) {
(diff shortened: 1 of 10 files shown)


yahoocla:
CLA is valid!

src-code:
Need to fix the tests, and then add some docs....

src-code:
Still needs some doc updates, please don't merge until that's added

bin/atomizer
```diff
@@ -14,16 +14,30 @@ var path = require('path');
...
 var params = require('minimist')(process.argv.slice(2));
 
 if (process.argv.slice(2).length === 0 || params.help) {
-    var usage = ['usage:  ', process.title, '[-o|--outfile=<file>] [--help] configfile ...'].join(' ');
+    var usage = ['usage:  ', process.title, ' -c|--config=<file> [-o|--outfile=<file>] [--help] [--verbose] [<files-to-parse> ...]'].join(' ');
     console.log(usage);
     return;
 }
 
+function handleMergeArrays (a, b) {
```
renatoi: We should move this function to `utils` so we don't have duplicates.

src/atomizer.js
```diff
@@ -11,54 +11,279 @@ var Absurd = require('absurd');
...
+var verbose = false;
 
-    var content;
+/**
+ * helper function to handle merging array of objects
+ * @param  {mixed} a Data of the first merge param
+ * @param  {mixed} b Data of the second merge param
+ * @return {mixed}   The merged object
+ */
+function handleMergeArrays (a, b) {
```
renatoi: Just point out that we need to move this to `utils`.

renatoi:
Other that, I think it looks good to me. We should update the docs to reflect how to use, specially programatically.

One feedback though. As a developer, I'd like to have an option to parse a file and get the css, either through chaining it or one call.

```javascript
var atomizer = require('atomizer');

// either this
atomizer.parse(src).getConfig().createCSS();

// or this
atomizer.atomify(src);
```

src-code:
I don't know if chaining is realistic (since that'd require the APIs to change their return value, and maintain state), but I do like the idea of offering something like `atomize()` that wraps all these calls...

Issue #66, opened 2015-03-03
Add CLI options for generating default config


src-code:
Right now the CLI tool requires you to pass a default config file, because at the very least we need the `config` preamble at the beginning of the config file (stuff like `namespace`, `start`, `end` are defined within `config`.)  We should allow those config params to be passed as args to the CLI so that users won't have to manually create a config file to get started.

renatoi:
Maybe these should be optional as well, as we could have a default value for them.

```javascript
{
  namespace: '#atomic',
  start: 'left',
  end: 'right'
}
```

src-code:
@renatoi Yeah, definitely!

src-code:
`start`/`end` have been replaced by `rtl` option.  `namespace` is now also an option, but is not yet exposed via the CLI (issue #89)

PR #70, opened 2015-03-05
Changing how we do general configuration for Atomizer

src-code:
Here's some big changes to how we configure Atomizer.  The gist is:

* No longer passing a `config` section in the config file.  Now those values are all passed via `options` in the Atomizer lib.  The Atomizer lib has default values for these options, and the CLI will expose these options soon (issue #66)
* `start` and `end` options are removed in favor of a new option called `rtl` (a boolean).  Based on the value of `rtl`, we'll replace the start and end tokens appropriately.
* Removed unused `defaults` config/option

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "atomizer",
-  "version": "1.0.1",
+  "version": "2.0.0",
   "description": "A tool for creating Atomic CSS, a collection of single purpose styling units for maximum reuse",
   "main": "./lib/atomic.js",
   "contributors": [
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "atomizer",
-  "version": "2.0.0",
+  "version": "2.0.0-alpha.1",
   "description": "A tool for creating Atomic CSS, a collection of single purpose styling units for maximum reuse",
   "main": "./lib/atomic.js",
   "contributors": [
--- a/src/atomizer.js
+++ b/src/atomizer.js
@@ -260,9 +260,9 @@ module.exports = {
 
         var atomicBuilder = new AtomicBuilder(rules, config, options);
         var build = atomicBuilder.getBuild();
-        if (!_.size(build)) {
-            throw new Error('Failed to generate CSS. The `build` object is empty.');
-        }
+        // if (!_.size(build)) {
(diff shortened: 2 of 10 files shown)


yahoocla:
CLA is valid!

src/atomizer.js
```diff
@@ -235,7 +235,15 @@ module.exports = {
             minify: false,
             keepCamelCase: false,
             extCSS: '.css',
-            banner: ''
+            banner: '',
+            namespace: '#atomic',
+            rtl: false,
+            // TODO: Verify these are good defaults
```
src-code: Not sure if we want to have these as default values?  Thoughts?

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "atomizer",
-  "version": "1.0.1",
+  "version": "2.0.0",
```
renatoi: We can bump it now, but I'd hold the publish after we implement pseudo classes and breakpoints. Thoughts?

src/atomizer.js
renatoi: We might need to change how we deal with breakpoints since @thierryk mentioned the need for more complex use of media-queries. For now, I think it's fine to have defaults breakpoints.

src/atomizer.js
```diff
@@ -250,11 +258,11 @@ module.exports = {
             api.import(options.require);
         }
 
-        var atomicBuilder = new AtomicBuilder(rules, config);
+        var atomicBuilder = new AtomicBuilder(rules, config, options);
         var build = atomicBuilder.getBuild();
-        if (!_.size(build)) {
-            throw new Error('Failed to generate CSS. The `build` object is empty.');
-        }
+        // if (!_.size(build)) {
```
renatoi: I guess we can get this removed.

package.json
src-code: I could also just publish an alpha or beta release, so we can more easily play around with it as we develop new features.

src/atomizer.js
src-code: Hmmm, perhaps opening it up to allow arbitrary breakpoint naming?  Let's start tracking that in another issue.

src-code: Issue #71 tracks the breakpoint question.

src/atomizer.js
src-code: Will do.

PR #72, opened 2015-03-05
Adding mergeConfigs() API

src-code:
New method `mergeConfigs()`, which takes an array of configuration objects as its lone argument.  

fyi @renatoi


yahoocla:
CLA is valid!

renatoi:
LGTM

PR #74, opened 2015-03-06
Add border-radius for corners + border-color/width/style

thierryk:
@src-code @renatoi please check this as I'm not 100% sure of the proper use of `allowCustomAutoSuffix`.

Also, note that for yahoodotcom I have included a bunch of border rules as **helper classes** to make the solution less verbose when we are dealing with *single pixel* borders; meaning we'll be able to do things like this:

```html
<div class="Bdx-1 Bdc-000"></div>
```

Do you guys think we should include those helpers in this project too?

### DIFF
--- a/src/rules.js
+++ b/src/rules.js
@@ -110,6 +110,121 @@ module.exports = [
         allowCustom: true,
         allowSuffixToValue: true
     },
+    // top-right
+    {
+        type: 'pattern',
+        id: 'border-top-right-radius',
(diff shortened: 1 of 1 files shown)


yahoocla:
CLA is valid!

src-code:
@thierryk Makes sense to me to have rules for border width, if we already have rules for color and line style.  All in all, this LGTM!

renatoi:
LGTM

Issue #71, opened 2015-03-05
Flexible Breakpoint config


src-code:
@thierryk recently pointed out that Atomizer's breakpoint design, which limits you to three breakpoints, is not flexible enough to accommodate common use cases.  We probably should allow developers to create unlimited arbitrary breakpoints.  

Also, it'd be a good opportunity to revisit whether it makes sense to have any default breakpoints (such as the 3 we have now), and if so, what their default values should be.

@renatoi

renatoi:
Breakpoints idea in Atomizer was inspired by Bootstrap Grid System:
http://getbootstrap.com/css/#grid-options

Allowing ultra flexible breakPoints might seem a good idea but I wonder if that's ideal/good practice.

There are some problems:

1. Conflicting Breakpoints: If two components declare two different breakpoints with the same name but with different values, which one should be used? I think it makes more sense if the app responsible to lay out the grid is actually the solely owner of breakpoint declarations.

2. If a component declares its own unique breakpoints, then we have bloat. The more we go towards custom rules the more bloat we have. On the other hand, it sucks to limit how the component react to different viewport sizes (Ideally, it should react to its OWN size, not the viewport, but this is not doable with CSS alone at the moment and very costly to do it with JS). So we either limit the capability of breakpoints on purpose to enforce good practice or we allow custom breakpoints to be declared at the cost of bloat. Keep in mind that the bloat I'm talking about here is not only in the CSS, but also in the markup.

I honestly prefer to keep having 3 breakpoints defined by the top level component and if any child component needs to have its own set of custom rules, then they should create their own custom CSS (not atomic). At least the bloat is only in one place.

thierryk:
I think for now we should leave things they way they are and see if users open issues about this and what are their asks exactly. 

As a side note, I'm not really in favor of *enforcing* good practice. In my opinion, best practice should be a *suggestion*, but not be imposed onto the user. After all, there are exceptions to every rules...

renatoi:
@thierryk suggested creating an option for the breakpoints so the rules are contained in the media query.

```css
/**
 * computed if:
 * - no responsive class have been used OR
 * - it is not computed due to MQ conditions.
**/
.D-b {
    display: block;
}

/* computed once viewport is at least 300px */
/* if we grow the viewport, this rule will still be computed one to the element it is applied */
@media (min-width: 300px) {
    .D-b--sm {
        display: block;
    }
}

/* computed only if the viewport width is less than 300px */
[...]

PR #75, opened 2015-03-06
[WIP] Pseudos and Breakpoints

renatoi:
* Added support for pseudos and breakpoints in custom suffixes;
   * BreakPoints config now requires a media-query declaration instead of just a number (See issue #71);
   * Custom BreakPoints can be declared;

Still could not figure out a good way to do pseudos for non-custom suffixes.

```
{
    display: {
        b: true,
        'b:h': true
    }
}
```

`b:h` is the ideal here but it makes my search in the config inefficient since I'm checking if the rule exist just by looking at the hash table for a valid keyword (in this case `b`). Is there a good way to look for the key in the hash table with the variant `b:h`?

@src-code thoughts?

### DIFF
--- a/src/atomizer.js
+++ b/src/atomizer.js
@@ -253,9 +253,9 @@ module.exports = {
             rtl: false,
             // TODO: Verify these are good defaults
             breakPoints: {
-                'sm': '767px',
-                'md': '992px',
-                'lg': '1200px'
+                'sm': '@media(min-width:767px)',
+                'md': '@media(min-width:992px)',
+                'lg': '@media(min-width:1200px)'
             }
         }, options);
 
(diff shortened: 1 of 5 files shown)


yahoocla:
CLA is valid!

src-code:
@renatoi Does this one need to be broken up now that we've been considering more drastic changes to config to accommodate pseudos?

src-code:
This pull may be replaced by work being done as part of issue #90.

src/atomizer.js
```diff
@@ -253,9 +253,9 @@ module.exports = {
             rtl: false,
             // TODO: Verify these are good defaults
             breakPoints: {
-                'sm': '767px',
-                'md': '992px',
-                'lg': '1200px'
+                'sm': '@media(min-width:767px)',
+                'md': '@media(min-width:992px)',
```
3den: :+1:

src-code:
Closing in favor of pull #108

PR #77, opened 2015-03-09
Merge configs fix

renatoi:
@src-code 

mergeConfigs() should return an empty object if argument is not an array or if it's an empty array

### DIFF
--- a/src/atomizer.js
+++ b/src/atomizer.js
@@ -175,7 +175,7 @@ module.exports = {
      * @return {object} An atomizer config object
      */
     mergeConfigs: function (configs) {
-        return Array.isArray(configs) && configs.length > 0 ? _.merge.apply(null, configs.concat(utils.handleMergeArrays)) : {};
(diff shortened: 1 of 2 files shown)


yahoocla:
CLA is valid!

src/atomizer.js
```diff
@@ -175,7 +175,7 @@ module.exports = {
      * @return {object} An atomizer config object
      */
     mergeConfigs: function (configs) {
-        return _.merge.apply(null, configs.concat(utils.handleMergeArrays));
+        return Array.isArray(configs) && configs.length > 0 ? _.merge.apply(null, configs.concat(utils.handleMergeArrays)) : {};
```
src-code: Are we polyfilling this?  Only works in IE9+.  Otherwise we need to use lodash here, since this code could run client-side.

renatoi: I think we should. We should use http://polyester-site.rc.staging.manhattan.corp.bf1.yahoo.com/

renatoi: Yeah, nvm, using lodash now.

PR #78, opened 2015-03-09
Changing breakPoints to be configured via config rather than as options

src-code:
@renatoi As we discussed, breakPoints will now be configured via the `breakPoints` property on the config object rather than as options.  `atomizer.createCSS()` will continue to manage a default set of breakpoints however, which it will add to the config object if necessary.  In the future, we may want to enable alternative ways to specify breakpoints via the CLI options or grunt task options, but for now I'm leaving that TBD.

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,11 +43,21 @@ atomizer -c config.js > atomic.css
 var Atomizer = require('atomizer');
 
 var defaultConfig = {
+    "breakPoints": {
+        'sm': '750px',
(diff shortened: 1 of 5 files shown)


yahoocla:
CLA is valid!

renatoi:
LGTM. Make sure we merge PR #77 first so we can bump it all together.

src-code:
@renatoi Will do!  Thanks!

PR #84, opened 2015-03-11
Add rules for transition + swap axys for `Bgp-`

thierryk:
@src-code @renatoi I've added rules for transition but also swapped the letter in the suffix for background-position. I think it makes more sense to have the `x` axis first in there. No?

### DIFF
--- a/src/rules.js
+++ b/src/rules.js
@@ -551,10 +551,10 @@ module.exports = [
         prefix: '.Bgp-',
         properties: ['background-position'],
         rules: [
-            {suffix: 'ts', values: ['$START 0']},
-            {suffix: 'te', values: ['$END 0']},
-            {suffix: 'bs', values: ['$START 100%']},
-            {suffix: 'be', values: ['$END 100%']}
+            {suffix: 'st', values: ['$START 0']},
+            {suffix: 'et', values: ['$END 0']},
+            {suffix: 'sb', values: ['$START 100%']},
+            {suffix: 'eb', values: ['$END 100%']}
         ]
     },
     /* background-repeat */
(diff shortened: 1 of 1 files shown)


yahoocla:
CLA is valid!

src/rules.js
```diff
@@ -1903,6 +1903,80 @@ module.exports = [
...
+    },
+    {
+        type: 'pattern',
+        id: 'transition-timing-function',
+        name: 'Transition timing function',
+        prefix: '.Trstf-',
+        properties: ['transition-timing-function'],
+        allowCustom: true,
+        allowCustomAutoSuffix: false,
+        allowSuffixToValue: true,
```
src-code: Does this make sense?  I believe you could have custom values (eg for cubic-bezier), but you wouldn't be able to get this from the suffix

src/rules.js
```diff
@@ -1903,6 +1903,80 @@ module.exports = [
...
+    },
+    {
+        type: 'pattern',
+        id: 'transition-property',
+        name: 'Transition property',
+        prefix: '.Trsp-',
+        properties: ['transition-property'],
+        allowCustom: true,
+        allowCustomAutoSuffix: false,
+        allowSuffixToValue: true,
```
src-code: Does allowSuffixToValue make sense for this one?

src/rules.js
```diff
@@ -1903,6 +1903,80 @@ module.exports = [
...
+     */
+    {
+        type: 'pattern',
+        id: 'transition',
+        name: 'Transition',
+        prefix: '.Trs-',
+        properties: ['transition'],
+        allowCustom: true,
+        allowCustomAutoSuffix: false,
+        allowSuffixToValue: true
```
src-code: Thinking allowSuffixToValue doesn't make sense for this one...

thierryk: We could have something like `Trs-all_2s`isn't?

We need to make sure Atomizer replaces the `_` with a space though.
I'll open an issue to make sure we have that on our radar if this does not work like this yet.

src/rules.js
thierryk: We could use classes like `.Trsp-background`, `.Trsp-height`, or else.

src/rules.js
thierryk: Authors could use `Trstf-ease` or `Trstf-linear`. 

Not sure how Atomizer deals with suffix containing dashes (`-`); for example: `Trstf-ease-in-out`.

src/rules.js
src-code: I don't know, is that really the way we want to go with these suffixes?  Seems like it'd get unnecessarily complex very quickly - some of these take a lot of values...

src/rules.js
src-code: I don't understand why we'd want to support that - we already have rules defining `Trstf-e` and `Trstf-l` for ease and linear, and `Trstf-eio` for ease-in-out, so there's no reason to support suffix-to-value here.  Suffix-to-value should be reserved more for arbitrary measurement style values, I'd think.

src/rules.js
src-code: This still feels odd to me - I think of suffix-to-value being more for measurements and other numeric values.  Using strings like this as suffix-to-value could easily become confused with global suffixes and other custom suffixes that aren't values.

src/rules.js
thierryk: Yeah, I guess we can remove it then. I thought of it as a mean for people to use these values `as-is` in case they didn't remember/don't know the abbreviation for the said values, but it is true that we do not do that for other properties, so I'll switch to false then..

src/rules.js
thierryk: I think there is value in allowing people to do this because in any case *there is good chance that's the string they'd be using* to create their custom value. 
And we could namespace global suffixes to avoid any confusion, no?

src/rules.js
thierryk: I think people will use custom suffixes for complex values (i.e. multiple values separated by commas) but for the most common used case like the example I give above, people will want to simply pass a property and duration. Why forcing them to use that very same string as a custom suffix when Atomizer could do that for them *on the fly*?

src-code: I think because it makes the syntax less clear.  We'd have to support this for all rules that have `allowSuffixToValue`, and are we sure that won't introduce ambiguity or conflict in other styles' values?  For that reason, I think that if we want to go down this path, we need to carefully consider and test this change.

At any rate, for now this syntax doesn't work in Atomizer, so for the sake of your pull request, it should be disabled in  the rules file.  We can file an issue for the feature support and enable this flag later if support is added.

src/rules.js
src-code: True, but it's going to get ugly really quickly as some of these values could be quite verbose.  And for very short values, it could introduce conflicts with globals and custom suffixes that are unintended.  I don't think we should tackle this right now.

src-code:
LGTM.  Needs some conflict resolution so we can merge it...

PR #85, opened 2015-03-11
Add separation (`_`) between "Bgp-" values + remove some trailing commas

thierryk:


yahoocla:
CLA is valid!

src-code:
Seems fine to me :shipit:

Issue #73, opened 2015-03-05
Let's remove the keyword values for `font-size`


thierryk:
Even though they should belong there since those are valid values for this property, I think it'll confuse people more than it'll help them.
So let's make `sm`, `md`, and the like custom values...

src-code:
Done in PR #92

PR #92, opened 2015-03-12
Remove keywords (issue #73)

thierryk:
@yasteve @renatoi  I also removed `suffixType: 'numerical'` as I don't think this does anything, does it?

### DIFF
--- a/src/rules.js
+++ b/src/rules.js
@@ -1054,7 +1054,8 @@ module.exports = [
         properties: ['font-size'],
         allowCustom: true,
         allowCustomAutoSuffix: true,
-        allowSuffixToValue: true
+        allowSuffixToValue: true,
+        suffixType: 'numerical'
     },
 
     /**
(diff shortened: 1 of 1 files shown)


yahoocla:
CLA is valid!

src-code:
`numerical` was for auto-incrementing suffixes... so if true, they're: 1,2,3..., and if false: a,b,c... 
By default Atomizer treats this param as 'false', so removing it would change the current behavior.  Not sure we intend to keep supporting this feature, however. @renatoi?

src-code:
Addresses issue #73

thierryk:
I put it back as I could imagine people willing to use `Fz1`, `Fz-2`, etc. After all, these should be considered sizes too.

renatoi:
I think we should get rid of these since it can mess up with `allowSuffixToValue`, also encouraging a bad practice (because it can create issues).

Example: `Lh-1`, `Lh-2`: Are these valid values or a sequence?

Ideally, as a good practice, custom classes shouldn't have a numerical sequence.

src-code:
LGTM.

Issue #94, opened 2015-03-12
Clean up default options


src-code:
Options currently given defaults by atomizer.js:

```javascript
{
   require: [],
   morph: null,
   combineSelectors: true,
   minify: false,
   keepCamelCase: false,
   extCSS: '.css',
   banner: '',
   namespace: '#atomic',
   rtl: false
}
```

`banner` is used by atomizer.js

`morph` and `require` are used by AbsurdJS (via atomizer.js).

`rtl` and `namespace` are used by AtomicBuilder.js currently.  

`extCSS`, `combineSelectors`, `minify`, and `keepCamelCase` I believe are not being used, and can be removed.

PR #97, opened 2015-03-13
Removing unused options

src-code:
Addressing issue #94.

### DIFF
--- a/src/atomizer.js
+++ b/src/atomizer.js
@@ -250,10 +250,6 @@ module.exports = {
         options = objectAssign({}, {
             require: [],
             morph: null,
-            combineSelectors: true,
-            minify: false,
-            keepCamelCase: false,
-            extCSS: '.css',
             banner: '',
             namespace: '#atomic',
             rtl: false


yahoocla:
CLA is valid!

renatoi:
We don't want to keep the minify there?

src-code:
I don't think `minify` is actually doing anything right now.  So we could add it back if we added support.

renatoi:
Keep it out for now then. Add it back if we think it's needed. LGTM.

src-code:
@renatoi I'm wrong, minify *is* doing something.  Apparently it's an Absurd option.  Along with `keepCamelCase` and `combineSelectors`.  It looks like we're setting the same default values that Absurd uses... not sure what these options do, aside from `minify`, however.

renatoi:
I'd say let's keep them out. Ideally combineSelectors and keepCameCase should be default in Atomizer. Minify could be an option, but not terribly necessary.

src-code:
Ok, sounds good to me.

PR #98, opened 2015-03-17
allow default unit to be set

3den:
@thierryk @renatoi This PR should allow default unit formats to be set

### DIFF
--- a/tests/atomizer.js
+++ b/tests/atomizer.js
@@ -281,7 +281,7 @@ describe('getConfig()', function () {
 
     it('should merge the default config with the newly created config', function () {
         var config;
-        var classNames = ['C-07f', 'Bgcp-bb', 'Fz-3em'];
+        var classNames = ['Bd-1', 'Bd-2', 'C-07f', 'Bgcp-bb', 'Fz-3em'];
         var expectedConfig = {
             'background-clip': {
                 bb: true
--- a/tests/atomizer.js
+++ b/tests/atomizer.js
@@ -281,7 +281,7 @@ describe('getConfig()', function () {
 
     it('should merge the default config with the newly created config', function () {
         var config;
-        var classNames = ['Bd-1', 'Bd-2', 'C-07f', 'Bgcp-bb', 'Fz-3em'];
+        var classNames = ['C-07f', 'Bgcp-bb', 'Fz-3em'];
         var expectedConfig = {
             'background-clip': {
                 bb: true
@@ -306,6 +306,53 @@ describe('getConfig()', function () {
 
         config = atomizer.getConfig(classNames, defaultConfig);
         expect(config).to.deep.equal(atomizer.mergeConfigs([defaultConfig, expectedConfig]));
+    });
+
+    it("should append the default unit", function () {
+        var classNames = ['P-1', 'M-0.3', 'P-2%', 'C-333'];
+
+        var defaultConfig = {
+            defaultUnit: 'px',
+            margin: {
+                defaultUnit: 'em'
+            }
+        };
+
+        var expectedConfig = {
+            defaultUnit: 'px',
+            padding: {
+                custom: [
+                    {
+                        "suffix": "2%",
+                        "values": [ "2%" ]
+                    },
+                    {
+                        "suffix": "1",
+                        "values": [ "1px" ]
+                    }
+                ]
+            },
+            margin: {
+                defaultUnit: 'em',
+                custom: [
+                    {
+                        "suffix": "0.3",
+                        "values": [ "0.3em" ]
+                    }
+                ]
+            },
+            color: {
+                custom: [
+                    {
+                        "suffix": "333",
+                        "values": [ "#333" ]
+                    }
+                ]
+            }
+        };
+
+        var config = atomizer.getConfig(classNames, defaultConfig);
+        expect(config).to.deep.equal(expectedConfig);
     });
 });
 
--- a/src/atomizer.js
+++ b/src/atomizer.js
@@ -102,6 +102,33 @@ function getConfigRule(className, currentConfig, warnings) {
 }
 
 /**
+ * Used by handleCustomConfigRule to sanitize the suffix value
+ * @param  {string} suffix        The suffix of the class.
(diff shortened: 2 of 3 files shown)


yahoocla:
CLA is valid!

src-code:
It seems this would make it impossible to combine components based on different unit configurations.  Eg, if I use a React Carousel that sets a default unit of `px`, but my main project is using a default unit of `em`, then the classes used in each would be the same, but expect to render very differently.  I'm not sure the convenience of leaving off the units is worth the cost of breaking portability.

thierryk:
@src-code good point! But don't you think this option could be valuable for small projects where teams agree on what should be the default unit or for developers for whom "portability" is not a concern at all?

My concern would be more related to parsing, I'm wondering if the lack of unit could throw off the parser in some weird cases where the suffix could become ambiguous(?).

tests/atomizer.js
```diff
@@ -307,6 +307,53 @@ describe('getConfig()', function () {
...
         expect(config).to.deep.equal(atomizer.mergeConfigs([defaultConfig, expectedConfig]));
     });
+
+    it("should append the default unit", function () {
+        var classNames = ['P-1', 'M-0.3', 'P-2%', 'C-333'];
+
+        var defaultConfig = {
+            defaultUnit: 'px',
+            margin: {
+                defaultUnit: 'em'
```
3den: @src-code @thierryk I though about that same issue too, and as you can see it is possible to have different default units for different properties. In this tests you can see that the defaultUnit set on the top level is `px` but we are able to override it to use `em` for margin.

This setting will be very helpful to the finance project not only for the migration but also to make sure we have a consistent use units, I do not want to allow people to use `px` in some paddings and `em` in others I want all paddings on the app to use the same unit to keep the ui consistent.

3den:
@thierryk I can add more tests to cover any of you concerns. Please let me know what are the use cases that you thing this might break so I can prevent that.

Please notice that this is an optional config, if people don't want to use it there is no side effect, and all tests we have are passing.

tests/atomizer.js
```diff
@@ -307,6 +307,53 @@ describe('getConfig()', function () {
...
+            defaultUnit: 'px',
+            padding: {
+                custom: [
+                    {
+                        "suffix": "2%",
+                        "values": [ "2%" ]
+                    },
+                    {
+                        "suffix": "1",
+                        "values": [ "1px" ]
```
3den: if the unit is not passed for padding the defaultUnit is given

tests/atomizer.js
```diff
@@ -307,6 +307,53 @@ describe('getConfig()', function () {
...
+            }
+        };
+
+        var expectedConfig = {
+            defaultUnit: 'px',
+            padding: {
+                custom: [
+                    {
+                        "suffix": "2%",
+                        "values": [ "2%" ]
```
3den: if the unit was passed we apply it.

tests/atomizer.js
```diff
@@ -307,6 +307,53 @@ describe('getConfig()', function () {
...
+                        "values": [ "1px" ]
+                    }
+                ]
+            },
+            margin: {
+                defaultUnit: 'em',
+                custom: [
+                    {
+                        "suffix": "0.3",
+                        "values": [ "0.3em" ]
```
3den: for margin the default unit is `em`.

tests/atomizer.js
thierryk: @3den I don't really agree about the "finance migration" though. The goal is to *unify* our styling to allow sharing components (increase portability). So even if I see value in this change for devs *outside* of Yahoo (who are not concerned by portability and such), I do not think it should be used *inside* Yahoo unless it is set at the top-level config. In other words, if we were settings a *default* unit, it would be the *same for everybody*.

3den: I think that for react components the best practice is not to use this feature and keep units explicit. But for applications this feature is very useful to enforce consistency. and remove a bit of duplication.

thierryk: Also, facilitating the finance migration in that regard would simply require to parse the files and append `px` at the end of `padding`, `margin` classes and the like. Which I don't think would require a lot of effort.

3den: is not about effort only is more about duplication, why do we need to append `px` over and over if we already decided that all paddings should use `px`?

This feature is very similar to the custom breakpoints and as i said it is optional people don't have to use it if they don't want.

thierryk: > I think that for react components the best practice is not to use this feature and keep units explicit.

Yes, I'd agree to that.

thierryk:
@3den If authors believe units are mandatory/best practice on their end, then they could easily use numbers as custom suffixes, for example: 

```css
.M-100 {
    margin: 100%;
}
```

In other words, if people's config does not use this feature then they may/could come up with something like the above.

3den:
@thierryk I agree, we should have a doc explaining how this feature works, when it is recommended to use it and when it is not. 

For building reusable node_modules it is much better to be explicit about the units. But on apps this can help to enforce unit consistency and remove some duplication of code.

src-code:
Let's keep this conversation going - this is a pretty significant and potentially risky change, so I don't want to rush this one in for 2.0.  Let's make sure we have all the edge cases covered well.

3den:
@src-code sounds good, in the mean time let me know if there is any edge case that I can cover with tests this feature is something that my team would like to have.

thierryk:
@3den 

> this feature is something that my team would like to have.

As I hinted earlier, I'm not sure you guys want to go this route. If we want to build one Yahoo, then **we want everybody to be using the same tool/values/whatever**. We do not want Atomizer to facilitate fragmentation. 

Actually, one thing that was great about Stencil is the fact that it enforced some sort of common styles across the board.

3den:
@thierryk this feature would be a "nice to have" for us but we can live without it. The reason why I sent this PR was because after we talked yesterday I though that it was a feature that you also wanted and you asked me to send a PR for it.

thierryk:
@3den The point I've been trying to make here is that this would be a nice feature/option to have *but not one we should use internally*.

PR #102, opened 2015-03-19
CLI refactor

src-code:
Updating the CLI to work with the latest APIs being worked on in the `refactor` branch.  Also fixed a couple of small issues in `atomizer.js` in the process.

@renatoi fyi


yahoocla:
CLA is valid!

PR #103, opened 2015-03-19
Add animation and transform

thierryk:
@src-code @renatoi 

I added `animation` and its variations but didn't do much regarding `transform` because of its nature -- since we need to pass functions and values (i.e. `rotate(90deg)`).

We talked about using aliases for the transforms too, I'll make sure we have an issue opened for that.

### DIFF
--- a/src/rules.js
+++ b/src/rules.js
@@ -14,6 +14,132 @@
  **/
 
 module.exports = [
+/**
+ ==================================================================
+ ANIMATION
+ ==================================================================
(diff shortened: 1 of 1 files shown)


yahoocla:
CLA is valid!

src-code:
We'll have to update this one for the rule syntax changes coming in the `refactor` branch.

thierryk:
Should we merge for now and revisit?

PR #104, opened 2015-03-20
Restoring namespace support

src-code:
Bringing back namespace support, this time to core Atomizer

@renatoi


yahoocla:
CLA is valid!

PR #105, opened 2015-03-20
Restoring Namespace support

src-code:
Bringing back namespace support, which was previously in AtomicBuilder

@renatoi


yahoocla:
CLA is valid!

PR #109, opened 2015-03-20
Updating tests for 100% coverage on reactors branch

src-code:
Lots of new tests, plus fixes for hex and signed (eg `neg`) values.  

Please review regex changes in particular carefully.  I used a negated lookahead regex for hex values in order to ensure it does not match if the digits are followed by a unit, otherwise it would incorrectly think things like `P-100px` or `H-100%` were hex values.


yahoocla:
CLA is valid!

PR #110, opened 2015-03-21
Refactor helpers

renatoi:
@src-code


yahoocla:
CLA is valid!

src/atomizer.js
```diff
@@ -500,28 +544,74 @@ Atomizer.prototype.getCss = function (classNames/*:string[]*/, config/*:Atomizer
...
                         }
-                    });
+                    }
+                    // a custom class name not declared in the config might not have values
+                    else if (treeo.value) {
+                        rule.properties.forEach(function (property) {
+                            var value = Atomizer.replaceConstants(treeo.value, options.rtl);
+                            property = Atomizer.replaceConstants(property, options.rtl);
+                            if (breakPoint) {
+                                csso[className][breakPoint][property] = value;
```
src-code: Is the `replaceConstants()` call missing here?

src/atomizer.js
```diff
@@ -500,28 +544,74 @@ Atomizer.prototype.getCss = function (classNames/*:string[]*/, config/*:Atomizer
...
-                    rule.properties.forEach(function (property) {
-                        var value = Atomizer.replaceConstants(treeo.value, options.rtl);
-                        property = Atomizer.replaceConstants(property, options.rtl);
+
+                    // we have params in declaration
+                    if (treeo.params) {
+                        treeo.params.forEach(function (param, index) {
+                            if (breakPoint) {
+                                for (var prop in cssoHelpers[className][breakPoint]) {
+                                    cssoHelpers[className][breakPoint][prop] = cssoHelpers[className][breakPoint][prop].replace('$' + index, param);
```
src-code: Does this already have RTL constants replaced?

renatoi: I was thinking about this one. Don't you think it's better to do this in the end on top of `content` before returning it? Because I feel we're calling this method everywhere, for Helpers specifically, I'll have to loop every key in `declarations` which I already don't like to do with params right now.

src-code: Actually yeah, not sure why I didn't do that in the first place.  The only risk would be if someone defined a custom that's called "$START" or "$END" I guess?

renatoi: I think that's fine, since those are keywords.

src-code: I wonder, should we change the `$START` and `$END` tokens to something more unique - eg, something that can't be used in a classname, so there's no chance of conflict?

renatoi: Yeah, that's safer. We could change to `__START__` and `__END__`, maybe?

src-code: Ok, I'll open a new issue/pull to consider alternative tokens!

src-code: I've opened issue #113 to track changes to RTL tokens

Issue #90, opened 2015-03-12
New configuration format


src-code:
The 1.x  configuration format is overly complex and verbose, and no longer fits well with how we're using Atomizer to parse classes.  Instead, we'll migrate 2.x to use a new config format that is class-centric, and then @renatoi will refactor/replace `AtomicBuilder.js` to no longer generate the 1.x config, but instead dump an array of classes into the config output, and then atomizer.js will parse those classes as necessary when generating the CSS.  (Basically, we're cutting out the middle stage of configuration and going straight from classes to CSS.)  

Custom classes/suffixes will continue to be supported in configuration as in 1.x.

PR #111, opened 2015-03-23
Breaking out config merge into new getConfig() method

src-code:
Creating a new method called `getConfig()` so we can more easily build tools which output the complete config used to generate the CSS.

fyi @renatoi

### DIFF
--- a/bin/atomizer
+++ b/bin/atomizer
@@ -102,8 +102,11 @@ if (filesToParse.length) {
     classnames = parseFiles(filesToParse, !!params.R);
 }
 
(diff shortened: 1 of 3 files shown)


yahoocla:
CLA is valid!

src-code:
In the process of making this change, we discovered some regex bugs, which I ended up pushing to this branch.  Guess we could separate them out into their own pulls, but might be easier to just leave them here since it'll all end up in pull #90 anyway.

renatoi:
LGTM

PR #114, opened 2015-03-24
Updating examples

src-code:
Updating with new config syntax


yahoocla:
CLA is valid!

src-code:
Note, this example exposes a bug in how `!important` is applied to rules that don't use suffix values: issue #115

renatoi:
LGTM

examples/example-config.js
```diff
@@ -1,424 +1,44 @@
...
-        start: true,
-        end: true,
-        c: true,
-        j: true,
-        m: true
-    },
-
-    // rule
-    'bfc': true
+    'custom': {
```
redonkulus: Can we add a comment about what these are?

examples/example-config.js
```diff
@@ -1,424 +1,44 @@
...
+    'classNames': [
+        // normal
+        'Td-u',
+        // pseudos
+        'Td-u:h',
+        // breakpoints
+        'Td-u--sm',
+        // custom
+        'H-uh',
+        'C-primary',
```
redonkulus: Add a comment that maps "primary" to the custom variables above.

renatoi:
@redonkulus @src-code I added more descriptive comments about what each thing does. let me know if that's too much or incorrectly written.

redonkulus:
:shipit:

PR #117, opened 2015-03-24
Updating README for 2.0.0-beta

src-code:
Updates for 2.0.0-beta

@renatoi

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,33 +44,33 @@ var Atomizer = require('atomizer');
 
 var defaultConfig = {
     "breakPoints": {
-        'sm': '750px',
-        'md': '1000px',
-        'lg': '1200px'
+        'sm': '@media(min-width=750px)',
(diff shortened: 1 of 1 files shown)


yahoocla:
CLA is valid!

renatoi:
I think we should add a link to grunt-atomizer somewhere in the README.

renatoi:
Other than that, LGTM.

src-code:
@renatoi Do you propose to link to any and all Atomizer-related packages that come along?  Eg, including 3rd-party authored projects, for instance if someone were to create plugins for other task runners?

renatoi:
@src-code Yeah, I think it makes sense to have a link to all related tools for Atomizer.

src-code:
Ok, let me add links in a new pull, so we don't have to hold up 2.0.0 any further!

PR #119, opened 2015-03-25
Added ability to pass prop+named in custom

renatoi:
@src-code @redonkulus @thierryk Now you'll be able to pass the prop + value identifier in ' custom', to contain the custom class to the property.

```
{
    custom: {
        'C-brand-color': '#400090',
        'Bgc-brand-color': '#000000'
    }
}
```

Note that it's not the class name you pass, but the property identifier + value identifier. This is important because if you use `C-brand-color:h`, you shouldn't pass that to custom, but `C-brand-color` only.

### DIFF
--- a/src/atomizer.js
+++ b/src/atomizer.js
@@ -173,7 +173,7 @@ interface AtomicTreeValue {
 /**
  * constructor
  */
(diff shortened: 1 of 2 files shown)


yahoocla:
CLA is valid!

src-code:
LGTM!

Issue #100, opened 2015-03-17
Syntax for rgba() color


thierryk:
I think we could go this route:

```css
.C-FFF\.4 {
    color: rgba(255, 255, 255, .4); 
}
```

Same as we have now, `C-` followed by `hex` value to what we'd add the transparency value.
The `.` in there can be used as marker since `1` (opaque) should be expressed as plain `grb()`/`hex` value.
Thoughts?

src-code:
Makes sense to me, though you'd have to escape that `.` so it isn't confused for a combined selector.

thierryk:
Good catch! That was a typo. I added the escape.

3den:
:+1:

src-code:
@thierryk Can we target the 2.1 milestone for this one?  I'm thinking 2.1 would be a fast-follow on 2.0 - I just want to avoid creeping on 2.0 and delaying it any further...

thierryk:
This is something we've integrated in Stream, we can keep the rule in `patch.css` until 2.1 is released but I'm not sure it is not something we may need asap as people start to "atomize" their component and I'm not sure how colors are specced there.

src-code:
@renatoi has this working already, so we'll sneak it in for 2.0

PR #121, opened 2015-03-25
Hex changes

renatoi:
* Breaking change: Hex syntax changed to include a '#' in the class name
   * Added support for alpha transparency. Example: '#333.4' or '#333333.4'


yahoocla:
CLA is valid!

src/atomizer.js
```diff
@@ -77,7 +77,7 @@ var GRAMMAR = {
     'SIGN'       : 'neg',
     'NUMBER'     : '[0-9]+(?:\\.[0-9]+)?',
     'UNIT'       : '[a-zA-Z%]+',
-    'HEX'        : '[0-9a-f]{3}(?:[0-9a-f]{3})?',
+    'HEX'        : '[0-9A-F]{3}(?:[0-9A-F]{3})?',
```
src-code: I'm not a big fan of uppercase hex values (feels like something I'd never do in CSS).  One way to avoid collisions is to just use 6 chars instead of 3 in these cases.  Or, we could introduce hex syntax that is more explicit (eg, `C-#333`), which would help us when dealing with numeric custom suffixes too (eg, 'C-100' should be valid as a custom suffix, and not be misinterpreted for a hex value, but today it's not possible.)

thierryk:
I don't have a problem with uppercase for `hex` values but I'd prefer `C-#333` than `C-333333`.

renatoi:
This is a breaking change. We won't bump major since we're in prerelease.

renatoi:
@src-code @thierryk Ok, I changed the syntax of hex to include '#' and they need to be lower case. It also accepts alpha numeric value (2 digits MAX) and it will be converted to RGBA.

src-code:
Addresses issues #122 and #100

src/atomizer.js
```diff
@@ -77,7 +78,8 @@ var GRAMMAR = {
     'SIGN'       : 'neg',
     'NUMBER'     : '[0-9]+(?:\\.[0-9]+)?',
     'UNIT'       : '[a-zA-Z%]+',
-    'HEX'        : '[0-9a-f]{3}(?:[0-9a-f]{3})?',
+    'HEX'        : '#[0-9a-f]{3}(?:[0-9a-f]{3})?',
+    'ALPHA'      : '\\.\\d\\d?',
```
src-code: Better to do `'\\.\\d{1,2}'` perhaps?

src-code:
LGTM

thierryk:
> It also accepts alpha numeric value (2 digits MAX) 

What do you mean by 2 digits? 2 digits after the `.` right? As in `C-#fff.33` ?

renatoi:
@thierryk Yes, maximum of 2 digits. Anything else will be ignored.

thierryk:
@renatoi and why do we enforce lowercase?

renatoi:
@thierryk for the reasons @src-code stated ("feels like something I'd never do in CSS" - even though it is accepted). Also, to enforce one way of using it to avoid having duplicate class names that points to the same value: '.C-\#FFF, C-\#fff { color: #fff }`.

thierryk:
> for the reasons @src-code stated ("feels like something I'd never do in CSS

That can't be because this is one of the most controversial stuff that comes up when we write CSS style guides. Some people say it should be uppercase, other say it should be uppercase. It's about 50/50.

> '.C-#FFF, C-#fff { color: #fff }`

That one makes a lot of sense. Thanks.

PR #132, opened 2015-03-27
Regex Fixes

renatoi:
@src-code @thierryk 

Fixed a problem with:

   * Helpers param Regex (I'm not proud of the solution though)
   * Parent selectors
   * Added sibling combinator
   * Added ability to have exclusive decimal number value. (e.g. `Trsdu-.3s` and `Op-.3`

### DIFF
--- a/src/atomizer.js
+++ b/src/atomizer.js
@@ -71,7 +71,7 @@ PSEUDO_REGEX = '(?:' + PSEUDO_REGEX.join('|') + ')';
 // regular grammar to match valid atomic classes
 var GRAMMAR = {
     'BOUNDARY'   : '(?:^|\\s|"|\'|\{)',
-    'PARENT'     : '[^\\s:>_+\'"`]+',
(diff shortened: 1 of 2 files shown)


yahoocla:
CLA is valid!

src-code:
@renatoi What part of the helpers param regex aren't you happy with?

renatoi:
@src-code It's actually a fix I had to include for the parent selector:

```
'PARENT'     : '[^\\s:>_\'"`]+',
```
I had to include `"`, `'` and `` ` `` to the negative character class. Basically, find any character that is not a space, colon, a parent separator, double quotes, single quotes, backtick. Do you have a better idea to match the parent selector?

renatoi:
Maybe I shouldn't be that open and go more restrictive, making an assumption of the possible characters for a parent class name?

`[a-zA-Z][-_a-zA-Z0-9]+?`

src-code:
@renatoi Yeah, I kind of like the idea of whitelisting instead of blacklisting there

PR #134, opened 2015-03-27
Fix pseudo class matching

ktslwy:
Found an issue when using certain pseudo classes. For example, in the pseudo class regex the `:l` comes before the `:li` and the regex doesn't check if there are more characters after. So using `:li` gets matched with `:left`. Adding a negative lookahead check to make sure the full string is read. The test case can reproduce the issue without the fix.


yahoocla:

ktslwy:
CLA signed

yahoocla:
CLA is valid!

renatoi:
Thanks! Looks good to me.

PR #135, opened 2015-03-30
Add 'transparent' and 'inherit' to background-color

thierryk:
@renatoi @src-code I just added missing values for `background-color`.


yahoocla:
CLA is valid!

PR #137, opened 2015-03-31
Add text overflow property

thierryk:
"text-overflow" was missing


yahoocla:
CLA is valid!

src/rules.js
```diff
@@ -1841,6 +1841,39 @@ module.exports = [
...
+     TEXT-OVERFLOW
+     ==================================================================
+     */
+    {
+        type: 'pattern',
+        name: 'Text overflow',
+        prefix: 'Tov-',
+        properties: ['text-overflow'],
+        allowCustom: true,
+        allowSuffixToValue: false,
```
src-code: I guess technically you should be able to pass a custom string as your value, right?  But that could get ugly with parsing...

src/rules.js
```diff
@@ -1841,6 +1841,39 @@ module.exports = [
...
+        rules: [
+            {suffix: 'c', values: ['clip']},
```
src-code: Any thoughts on adding support for left overflow?  (eg, `text-overfow: clip ellipses`, or `text-overflow: clip clip`, etc)

src/rules.js
thierryk: People should be able to do that using a custom value, but I preferred to set `allowSuffixToValue` to `false` to prevent potential parsing issues.

src/rules.js
thierryk: People should be able to do that via a custom value for that property. If I hadn't disable `allowSuffixToValue` they could even pass it as the value identifier, for example: `Tov-c_e`, but I think it's too dangerous.

src-code:
Ok, LGTM.

PR #139, opened 2015-04-01
Add values for "cursor"

thierryk:
I made the decision to omit "alias" as it kinda conflicts with "auto"


yahoocla:
CLA is valid!

src-code:
Seems right to exclude `alias`.  LGTM.

PR #141, opened 2015-04-01
Follow the 2 letters syntax for values

thierryk:


yahoocla:
CLA is valid!

renatoi:
You will review all value identifiers we have today, right? So I guess we will either see a new PR or you update this one.

thierryk:
yes, but I don't remember if we said we were going for the 1st letter of syllables or only fixing things where we don't have 2 letters for separated words value.

src-code:
@thierryk I'd think the latter, but what might be some examples of the former?

thierryk:
@src-code my last commit is to follow the "1st letter for each word sepearted bu a `-`" pattern. I really think we should leave it at that because the Emmet syntax follows some arbitrary syntax. 
For example we have this:
```javascript
{suffix: 'se', values: ['step-send']},
{suffix: 'ss', values: ['step-start']}
```

or even

```javascript
{suffix: 'tbcl', values: ['table-column']},
{suffix: 'tbclg', values: ['table-column-group']},
{suffix: 'tbhg', values: ['table-header-group']},
{suffix: 'tbfg', values: ['table-footer-group']},
{suffix: 'tbrg', values: ['table-row-group']}
```

Emmet made a few calls such as:

   * `z` stands for `size`
   * `tb` stands for `table` no matter where it is found
   * `ma` stands for `max` and `mi` for `min`. 

With the latter, they do not follow the pattern we adopted. For example, they do not use `mc` for `max-content` but `maxc`. Which makes sense since it avoids conflict with `min-content` in a nicer way than we would (`mc` and `mic`).

src-code:
LGTM.

PR #144, opened 2015-04-02
Fix #143

### LINKED ISSUE #143: LineClamp output is incorrect
It's been noted that when you have multiple instances of a helper like `LineClamp`, you may get incorrect CSS output such as:

```css
.LineClamp\(2\,32px\),.LineClamp\(2\,39px\),.LineClamp\(3\,50\.7px\) {
    -webkit-line-clamp: 2;
    max-height: 32px
}
```
I suspect this is somehow related to the logic used to combine selectors, possibly due to use of parens, but that's just a hunch.  Will need to look more deeply into this.

@renatoi @thierryk @karenrfay @kfay

renatoi:
@src-code @thierryk

### DIFF
--- a/src/atomizer.js
+++ b/src/atomizer.js
@@ -625,9 +625,9 @@ Atomizer.prototype.getCss = function (config/*:AtomizerConfig*/, options/*:CSSOp
                     }
 
                     if (breakPoint) {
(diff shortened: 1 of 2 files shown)


yahoocla:
CLA is valid!

PR #147, opened 2015-04-03
No namespace for contextual selectors

renatoi:
@thierryk @src-code

### DIFF
--- a/src/atomizer.js
+++ b/src/atomizer.js
@@ -446,6 +446,8 @@ Atomizer.prototype.getCss = function (config/*:AtomizerConfig*/, options/*:CSSOp
         }
         if (match.parent) {
             treeo.parent = match.parent;
+            // enforce !important, since these classes don't have the namespace
+            match.important = true;
         }
         if (match.parentPseudo) {
             treeo.parentPseudo = match.parentPseudo;
@@ -611,8 +613,12 @@ Atomizer.prototype.getCss = function (config/*:AtomizerConfig*/, options/*:CSSOp
                 // add the dot for the class
                 className = ['.', className].join('');
 
-                // finaly, create the object
+                // add the namespace
+                if (options.namespace && !treeo.parent && rule.type !== 'helper') {
+                    className = [options.namespace, ' ', className].join('');
+                }
 
+                // finaly, create the object
                 // helper rules doesn't have the same format as patterns
                 if (rule.type === 'helper') {
                     jssHelpers[className] = {};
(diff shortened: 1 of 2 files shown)


yahoocla:
CLA is valid!

src/atomizer.js
```diff
@@ -678,11 +684,6 @@ Atomizer.prototype.getCss = function (config/*:AtomizerConfig*/, options/*:CSSOp
         }
     });
 
-    if (options.namespace) {
-        var jssNew = {};
-        jssNew[options.namespace] = jss;
-        jss = jssNew;
-    }
     if (options.helpersNS) {
```
src-code: So Helpers always get namespaced?  Even if the Helper is part of a contextual selector?

renatoi: Yes, for now.

thierryk: For now as in "this will change" right? Because things will break for the same reason...

renatoi: Helpers are not namespaced today by default, only if you pass a namespace.

thierryk: I think we decided long ago to **not** namespace helpers, but your reply to @src-code confused me; because you answered:

> Yes, for now.

to

> So Helpers always get namespaced? Even if the Helper is part of a contextual selector?

which to me suggests that helpers **do** get namespaced, no?

but then what does this mean exactly?

> Helpers are not namespaced today by default, only if you pass a namespace.

Can we pass a namespace to helpers only?
I'm confused...

src-code: @thierryk Currently namespaces on helpers is optional, with the default being disabled.  What I'm more concerned by is, if you do specify a namespace for helpers, shouldn't we skip the namespace if a contextual selector is used, just like we do for atomic classes?

src-code: On a related note, I really think we need to rename the `helpersNS` option.  It's not consistently named compared to `namespace` for Atomic classes.  I'd rather we call it something like `namespace_helpers` (so it's visually grouped with `namespace`) or I suppose `helpers_namespace`.

renatoi: @thierryk Helpers can be namespaced if you pass a namespace to it. We talked about this before. The main reason is because of base contextualization, if that happens, then you might want to bump helper classes as well.

src-code:
LGTM, though there's some duplication between helpers and atomic processing that could probably be unified now that we're going to have the same syntax for each type of classes (for multiple params, etc)

PR #151, opened 2015-04-08
Removing dot from prefixes

src-code:
A few stray dots remained after being previously removed... @renatoi @thierryk fyi


yahoocla:
CLA is valid!

renatoi:
LGTM

PR #156, opened 2015-04-13
Being more defensive when no classnames are passed

src-code:
Handling this edge case a little more gracefully...


yahoocla:
CLA is valid!

renatoi:
LGTM

PR #159, opened 2015-04-14
parent-namespaced calsses with predefined values should have important

renatoi:
@src-code


yahoocla:
CLA is valid!

gbor:
@renatoi looks good to me.  Thanks for fixing this bug. =)

src-code:
LGTM.

PR #160, opened 2015-04-14
Fixes #158

### LINKED ISSUE #158: $ symbol in custom won't be generated in CSS
In the following case, it won't generate,
```js
{
  "custom": {
    "$uhHeight": "200px"
  }
} 
```
But if put another ``uhHeight``, both will be generated.

renatoi:
@src-code @hankhsiao


yahoocla:
CLA is valid!

PR #163, opened 2015-04-14
Fix LineClamp (issue #161)

### LINKED ISSUE #161: LineClamp should use display: flex
@thierryk

thierryk:
@renatoi as discussed, I added `display:flex` to 1 of the LineClamp rules to fix #161


yahoocla:
CLA is valid!

PR #164, opened 2015-04-15
Making 2nd arg to parseConfig() optional

src-code:
@renatoi @thierryk 
This makes it easier to use the `parseConfig()` API outside of `getCss()` since we may not always need to pass options


yahoocla:
CLA is valid!

renatoi:
LGTM

PR #169, opened 2015-04-17
Declarations

renatoi:
@src-code @thierryk

### DIFF
--- a/src/atomizer.js
+++ b/src/atomizer.js
@@ -139,17 +139,17 @@ function Atomizer(options/*:AtomizerOptions*/, rules/*:AtomizerRules*/) {
 Atomizer.prototype.addRules = function(rules/*:AtomizerRules*/)/*:void*/ {
 
     rules.forEach(function (rule) {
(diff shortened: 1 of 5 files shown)


yahoocla:
CLA is valid!

src/atomizer.js
```diff
@@ -587,67 +395,45 @@ Atomizer.prototype.parseConfig = function (config/*:AtomizerConfig*/, options/*:
...
-                                        }
-                                    });
-                                });
-                                namedFound = true;
-                                return true;
-                            }
-                        });
+                    // first check if the named value matches any of the values
+                    // registered in arguments.
+                    if (rule.arguments && index <= rule.arguments.length && Object.keys(rule.arguments[index]).indexOf(matchVal.named) >= 0) {
```
src-code: `Object.keys` support is limited to IE9+ client side, so if we want to support IE7-8, we'll want to use lodash here.

src-code:
LGTM.

3den:
awesome :+1:

PR #172, opened 2015-04-17
Added comments back and added aliases

renatoi:
@src-code @thierryk

### DIFF
--- a/src/rules.js
+++ b/src/rules.js
@@ -19,7 +19,7 @@
  * NOTES:
  *
  * Depending on the selector you use to namespace these rules (id versus class),
- * their style weight will be either 0.1.1.0 or 0.0.2.0. We suggest using an id
+ * their style weight will be either 0,1,1,0 or 0,0,2,0. We suggest using an id
  * for the extra specificity.
  *
  * - look for top/right/bottom/left rules in the "offset" section.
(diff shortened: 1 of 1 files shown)


yahoocla:
CLA is valid!

src/rules.js
```diff
@@ -1699,6 +2163,289 @@ module.exports = [
...
+        }
+    },
+    // matrix3d
+    {
+        "type": "pattern",
+        "name": "Matrix 3d (transform)",
+        "matcher": "Matrix3d",
+        "allowParamToValue": true,
+        "styles": {
+            "transform": "matrix($0,$1,$2,$3,$4,$5,$6,$7,$8,$9,$10,$11,$12,$13,$14,$15)"
```
src-code: This one will probably look funny in the reference docs... you'll get 15 "value" strings in a row :-/

src-code:
LGTM, but @thierryk should weigh in before we merge this

src/rules.js
renatoi: It's 16! :) I was wondering if I should remove matrix.

thierryk:
Sorry @renatoi, you need to revert the [specificity thing] (https://github.com/yahoo/atomizer/commit/02f9ddb68d43810e727d1453d0e66aa4661c4633) change. My bad...

**After that**, LGTM, :shipit:

PR #173, opened 2015-04-18
Added helpers to Atomizer

renatoi:
@thierryk


yahoocla:
CLA is valid!

thierryk:
Duh! I didn't even notice the color diff - must have been tired...

Anyway, we have a problem with `SpaceBetween` because we do **not** reset the `line-height` for the inner boxes.

I think best would be to pass a param that would take care of this: `SpaceBetween(1.4)`

Thoughts?

As a side note, `SpaceBetween` works on acss.io but that's because we have a plug in `base.css` to style the children of the box with the same `line-height` we have on `<body>`.

renatoi:
As discussed, SpaceBetween will be out for now due to the extra complexity of using it, we can reconsider adding later if needed.

PR #175, opened 2015-04-21
Add missing quotes around value of content property

thierryk:
@renatoi Karen spotted the issue of the missing pseudo-element this morning (she got "..." in the middle of strings).


yahoocla:
CLA is valid!

kfay:
@thierryk it looks like there are a few more instances of "content" strings that probably also need this fix in this same file. (e.g. the one for `Ell:after`)

thierryk:
Thanks @kfay &mdash; I took care of `Ell` and `Cf` too.

kfay:
LGTM, thanks @thierryk !

PR #176, opened 2015-04-23
Bgz should accept custom values

skippykawakami:
Bgz wasn't allowing custom values (45px for example) while previous versions did

### DIFF
--- a/src/rules.js
+++ b/src/rules.js
@@ -649,7 +649,7 @@ module.exports = [
         "type": "pattern",
         "name": "Background size",
         "matcher": "Bgz",
-        "allowParamToValue": false,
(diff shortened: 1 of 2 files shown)


yahoocla:

yahoocla:
CLA is valid!

thierryk:
Thanks Mark!

As a side note, this will only allow single value syntax (i.e. `background-size:50%`). Because styles relying on multiple backgrounds syntax must be set via the config (custom classes) as we *do not allow* multiple params for the same declaration (i.e. `background-size:50px,100px`). For now, params are only used to pass values to different properties (i.e. `LineClamp(2,54px)`).

PR #182, opened 2015-04-28
Fixes #179

### LINKED ISSUE #179: Set different values to a variable depending on breakpoints
Imagine this:

```javascript
'custom': {
    '$columnWidth': '20px',
    '$gutter': {
        'sm': '10px',
        'md': '15px',
        'lg': '20px'
    }
}
``` 
Then in the markup:
```html
<div class="P($gutter)"></div>
```

The **same class** in the markup will create a different padding around the box depending on breakpoints. 
And changes are made via the config object &mdash; not via markup (same as if we were editing a rule in a style sheet).

renatoi:
@hankhsiao @thierryk

### DIFF
--- a/src/atomizer.js
+++ b/src/atomizer.js
@@ -287,7 +287,35 @@ Atomizer.prototype.parseConfig = function (config/*:AtomizerConfig*/, options/*:
                         }
                     }
                     if (value !== null) {
-                        treeo.declarations[prop] = treeo.declarations[prop].replace('$' + index, value);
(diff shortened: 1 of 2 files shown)


yahoocla:
CLA is valid!

hankhsiao:
:+1:

hankhsiao:
This is only for breakpoints, do we have it for context?

renatoi:
We could also add it for context. I can implement that in a separate PR.

PR #183, opened 2015-04-28
Hotfix, values shouldn't be optional for atomic classes

renatoi:
@3den quick fix, values shouldn't be optional for non helper classes. All atomic classes require ().

### DIFF
--- a/src/lib/grammar.js
+++ b/src/lib/grammar.js
@@ -129,12 +129,13 @@ function buildRegex(map, keyEx, valEx) {
         '(?<' + keyEx + '>',
             keys,
         ')',
-        // value is optional
         '(?:\\(',
             '(?<' + valEx + '>',
                 GRAMMAR.VALUES,
             ')',
-        '\\))?',
+        '\\))',
+        // value is optional only for helper
+        keyEx === 'helper' ? '?' : ''
     ].join('');
 }
 
--- a/tests/atomizer.js
+++ b/tests/atomizer.js
@@ -240,6 +240,34 @@ describe('Atomizer()', function () {
             var result = atomizer.getCss(config, {ie: true});
             expect(result).to.equal(expected);
         });
+        it ('returns css if coliding helper and atomic rule is used at the same time', function () {
+            var atomizer = new Atomizer();
+            var config = {
+                custom: {
+                    '$custom': '1px solid #000'
+                },
+                classNames: ['Bd($custom)', 'Bd', 'BdT']
+            };
+            var expected = [
+                '.Bd\\(\\$custom\\) {',
+                '  border: 1px solid #000;',
+                '}',
+                '.Bd {',
+                '  border-width: 1px;',
+                '}',
+                '.Bd, .BdT {',
+                '  border-style: solid;',
+                '}',
+                '.BdT {',
+                '  border-top-width: 1px;',
+                '  border-right-width: 0;',
+                '  border-bottom-width: 0;',
+                '  border-left-width: 0;',
+                '}\n'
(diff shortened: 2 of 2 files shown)


yahoocla:
CLA is valid!

src/lib/grammar.js
```diff
@@ -129,12 +129,13 @@ function buildRegex(map, keyEx, valEx) {
...
         ')',
-        // value is optional
         '(?:\\(',
             '(?<' + valEx + '>',
                 GRAMMAR.VALUES,
             ')',
-        '\\))?',
+        '\\))',
+        // value is optional only for helper
+        keyEx === 'helper' ? '?' : ''
```
3den: :+1:  @renatoi why do we currently need to differentiate helpers from patterns?

renatoi: So ideally we wanted to have all share the same syntax. But based on many discussions, we made Helper classes not require the parenthesis. So you can use `Bd` and `BdT` alone. But this would only be for helper classes as atomic classes DO require a param no matter what since the pattern is always prop(value). Also atomic classes have very generic name like just `D` for display, making the value optional for this would make the tool find a lot of false positives. It sucks, I know. :(

3den: @renatoi we can implement so that the diference is based on the syntax of the matcher e.g.

```is
[
    {
        "name": "Border",
        // No args or parenthesis required
        "matcher": "Bd",
        "styles": {
            "border-width": "1px",
            "border-style": "solid"
        }
    },
  
    {
        "name": "Line clamp",
        // Parentesis with args optional 
        "matcher": "LineClamp(?)",
        "styles": {
            "-webkit-line-clamp": "$0",
            "max-height": "$1"
        },
        ...
    },

   {
        "name": "Animation",
        // Parenthesis with args required
        "matcher": "Anim($)",
        "allowParamToValue": true,
        "styles": {
            "animation": "$0"
        }
   }
]
```

Let me know what you think about this suggestion.

thierryk: @3den I don't think we would have optional parenthesis (i.e. `LineClamp`); we talked about this earlier and we decided against it.

3den: @thierryk that is fine the suggestion above would still work, in that case we just won't need to support the `(?)` suffix.

renatoi: I like the idea of adding a flag to tell when param is required or not, but I wouldn't touch the "matcher" part. I prefer adding another key to the object saying the param is optional.

renatoi: Actually, this is an issue today. LineClamp is required to have params but we are making it optional because it's a helper. So invalid LineClamp constructs will match and we don't want that. I'll try to get it fixed.

3den: :+1: it makes sense, we can use a boolean flag instead of a type

```js
[
    {
       // no param flag is required
        "name": "Border",
        "matcher": "Bd",
        "styles": {
            "border-width": "1px",
            "border-style": "solid"
        }
    },

    {
        "name": "Line clamp",
        "hasOptionalParams": true,
        "matcher": "LineClamp",
        "styles": {
            "-webkit-line-clamp": "$0",
            "max-height": "$1"
        },
        ...
    },

   {
        "name": "Animation",
        "hasParams": true,
        "matcher": "Anim",
        "allowParamToValue": true,
        "styles": {
            "animation": "$0"
        }
   }
]
```

renatoi: I'd have everything to have a param by default and a flag to tell when it's optional.

3den: thats cool

PR #187, opened 2015-04-30
Added ie opacity filter

### LINKED ISSUE #186: IE fallbacks
One fallback we noticed is missing is the IE filter opacity as an opacity fallback for IE8/7

How about font rem fallbacks etc?

renatoi:
Fixes #186 .


yahoocla:
CLA is valid!

hankhsiao:
:ship:

PR #189, opened 2015-05-02
default font values

3den:
@thierryk @renatoi https://github.com/yahoo/atomizer/issues/188

### DIFF
--- a/src/rules.js
+++ b/src/rules.js
@@ -1310,7 +1310,14 @@ module.exports = [
         "allowParamToValue": true,
         "styles": {
             "font-size": "$0"
-        }
+        },
+        "arguments": [{
+            "xs": "0.7rem",
+            "s": "0.8rem",
+            "m": "1rem",
+            "l": "1.5rem",
+            "xl": "2rem",
+        }]
     },
     /**
      ==================================================================


yahoocla:
CLA is valid!

src/rules.js
```diff
@@ -1310,7 +1310,14 @@ module.exports = [
         "allowParamToValue": true,
         "styles": {
             "font-size": "$0"
-        }
+        },
+        "arguments": [{
+            "xs": "0.7rem",
+            "s": "0.8rem",
```
3den: you guys can changes those values to ones that would make more sense in most apps

thierryk:
@3den We had that in an early version of Atomizer but we then decided against it. In our opinion, such values are tied to a Design Language; hence they belong to `config.js`, not to `rule.js`. 
Your example proves this actually as you are using `rem` unit which would make `Fz(xs)` bound to the styling of  `:root` (`<html>`).

Custom classes (from the `config`) are supposed to vary from project to project, but not the ones from `rules.js` - meaning adopting default (T-shirt size) values for `Fz()` would require people to either adopt opinionated values or overwrite them. And I think it would make little sense to overwrite Atomic classes... 

Thank you.

PR #190, opened 2015-05-02
Support multiple matrix value

roderickhsiao:
@renatoi @thierryk 

let me know if you think its fine :)

### DIFF
--- a/src/rules.js
+++ b/src/rules.js
@@ -2260,9 +2260,9 @@ module.exports = [
         "type": "pattern",
         "name": "Matrix (transform)",
         "matcher": "Matrix",
-        "allowParamToValue": true,
+        "allowParamToValue": false,
         "styles": {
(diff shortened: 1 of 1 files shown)


yahoocla:

src/rules.js
```diff
@@ -2272,7 +2272,7 @@ module.exports = [
         "matcher": "Matrix3d",
         "allowParamToValue": false,
         "styles": {
-            "transform": "matrix($0)"
+            "transform": "matrix($0,$1,$2,$3,$4,$5,$6,$7,$8,$9,$10,$11,$12,$13,$14,$15)"
```
renatoi: I originally coded like that as well but I realized this is an edge case and it's better to be defined in the config instead of being in the class name. Same to `matrix`. `allowParamToValue` should be false for both.

yahoocla:
CLA is valid!

renatoi:
Thanks @roderickhsiao

PR #193, opened 2015-05-06
Perf

renatoi:
@lingyan

### DIFF
--- a/src/atomizer.js
+++ b/src/atomizer.js
@@ -153,11 +153,20 @@ Atomizer.prototype.parseConfig = function (config/*:AtomizerConfig*/, options/*:
         }
 
         // check where this rule belongs to
+        // atomicSelector is the class name before the params: e.g. className(param)
+        // selector is the class name if params is not required
+        // we look both in rules and in helpers where this class belongs to
         if (this.rulesMap.hasOwnProperty(match.atomicSelector)) {
             ruleIndex = this.rulesMap[match.atomicSelector];
-        } else if (this.helpersMap.hasOwnProperty(match.atomicSelector)) {
+        }
+        // the atomicSelector can also be a helper that requires params
+        else if (this.helpersMap.hasOwnProperty(match.atomicSelector)) {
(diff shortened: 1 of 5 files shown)


yahoocla:
CLA is valid!

renatoi:
Some context:

Grammar is the lib we use to build the regex to parse atomic classes.

There are currently 3 regex that we're interested:

1. Find valid atomic class names on a text;
2. Parse the class name and make sense of it by storing the parsed data in an object;
3. Parse values inside of atomic classes;

For the the first case, we need a not-simple regex, it needs to include every single token of an atomic class name into the regex, so we only find valid classes.

For the second case, we don't need the regex to include everything. We're only interested in the syntactical structure, something like this:

`[parentclass] [:pseudo] [_] [ClassName] [(value)] [!] [:pseudo] [--breakPoint]`

For 3, the regex is simple and very fast since we're only parsing valid values.

Before I was using the same regex for 1 and 2. Now, I'm using a simple regex for 2, so it's faster. That's the `getSyntax(true)` ===> `true` means simple.

src/atomizer.js
```diff
@@ -141,18 +147,28 @@ Atomizer.prototype.parseConfig = function (config/*:AtomizerConfig*/, options/*:
...
+        if (this.rulesMap.hasOwnProperty(match.atomicSelector)) {
+            ruleIndex = this.rulesMap[match.atomicSelector];
+        } else if (this.helpersMap.hasOwnProperty(match.atomicSelector)) {
+            ruleIndex = this.helpersMap[match.atomicSelector];
+        } else if (this.helpersMap.hasOwnProperty(match.selector)) {
+            ruleIndex = this.helpersMap[match.selector];
+        } else {
+            // not a valid class, no op
+            return;
+        }
```
lingyan: Add more docs here would be helpful, like what is an `atomicSelector` and `selector`

lingyan:
Left a comment about doc.  Otherwise, :+1:

PR #194, opened 2015-05-18
cross browser user-select

3den:
@thierryk @renatoi this needed to make `user-select` work in all browsers.

### DIFF
--- a/src/rules.js
+++ b/src/rules.js
@@ -2526,7 +2526,12 @@ module.exports = [
         "matcher": "Us",
         "allowParamToValue": false,
         "styles": {
-            "user-select": "$0"
+            "user-select": "$0",
+            "-webkit-touch-callout": "$0",
+            "-webkit-user-select": "$0",
+            "-khtml-user-select": "$0",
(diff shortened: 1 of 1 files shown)


yahoocla:
CLA is valid!

renatoi:
We expect people to use [postcss auto-prefixer](https://github.com/postcss/autoprefixer) to prefix CSS properties after atomic css. Atomic CSS should remain pure CSS except for very minor hacks/cases if needed. This way we not only keep it clean but also adaptable to developers needs (I could only care only about 2 versions back while other could support 3 or more). So use autoprefixer with atomizer in your project.

3den:
@renatoi in that case can you simplify `grunt-atomizer` to do that automatically?

renatoi:
By design it is not the nature of this package to incorporate functionalities contained in other packages, specially with grunt, when there's already a [grunt package](https://github.com/nDmitry/grunt-postcss) that runs postcss after your css files. So just combine these grunt tasks to get the output you need.

PR #195, opened 2015-05-20
Allow numbers and letters [a-z0-9] on breakPoints

3den:
@renatoi @thierryk we recently needed this feature for creating more breakpoints.

We where using the tshirt size format but we wanted to be able to add more breakpoints (e.g. 2xs, xs, sm, m...).


yahoocla:
CLA is valid!

itssumitrai:
Looks good :+1:

renatoi:
Thanks @3den!

PR #207, opened 2015-06-03
allow param to value for transition-property

renatoi:


yahoocla:
CLA is valid!

koulmomo:
:+1: yes please!

PR #210, opened 2015-06-29
more specific styles should be defined after generic ones

3den:
X and Y styles should be defined before End, Start, Top and Bottom to allow overrides because they are more generic ones.

### DIFF
--- a/src/rules.js
+++ b/src/rules.js
@@ -190,6 +190,28 @@ module.exports = [
             }
         ]
     },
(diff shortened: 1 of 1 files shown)


yahoocla:
CLA is valid!

src-code:
LGTM, @renatoi any thoughts?

renatoi:
I agree. LGTM.

PR #211, opened 2015-06-30
Adding rules for font-smooth/font-smoothing

src-code:
Alright, we talked about this at one point and weren't sure we wanted to support a non-standard rule like `font-smooth`.  I've been running into some use cases lately where it's actually needed, and managing these rules apart from core Atomizer is a bit of a pain.  So I'm creating this PR to continue the discussion on whether Atomizer should have support baked in.  

@renatoi @thierryk Your thoughts?

### DIFF
--- a/src/rules.js
+++ b/src/rules.js
@@ -1301,6 +1301,65 @@ module.exports = [
         }
     },
     /**
+    ==================================================================
+    FONT-SMOOTH
+    Warning: This feature is non-standard and varies by UA. Use with caution.
+    https://developer.mozilla.org/en-US/docs/Web/CSS/font-smooth
+    ==================================================================
+    */
+    {
+        'type': 'pattern',
+        'id': 'font-smooth',
+        'name': 'Font smoothing',
+        'matcher': 'Fsm',
+        'allowParamToValue': true,
+        'styles': {
+            'font-smooth': '$0'
+        },
+        'arguments': [
+            {
+                'a': 'auto',
+                'n': 'never',
+                'aw': 'always'
+            }
+        ]
+    },
+    {
+        'type': 'pattern',
+        'id': 'font-smooth-webkit',
+        'name': 'Font smoothing - Webkit',
+        'matcher': 'Fsmw',
+        'allowParamToValue': false,
+        'styles': {
+            '-webkit-font-smoothing': '$0'
+        },
+        'arguments': [
+            {
+                'n': 'none',
+                'a': 'antialiased',
+                's': 'subpixel-antialiased'
+            }
+        ]
+    },
+    {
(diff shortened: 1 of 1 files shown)


yahoocla:
CLA is valid!

renatoi:
I am not a big fan of adding something that is non-standard. I believe base.css should contain the font-smooth in the body, no? Is there a use case that this is not being covered?

src-code:
@renatoi I've been seeing use cases where developers want to apply different values depending on the font used.  So applying in a base stylesheet doesn't work...

renatoi:
1. I think it would be bad to use something like `<div class="Fsm(aw) Fsmw(a) Fsmm(g)">` in the markup to support font-smoothing. Vendor-specific variations directly in the markup sounds bad. I'd rather use one class to do font-smoothing, it usually comes with the combo:

```css
.Fsmaw {
    font-smoothing: always;
    -webkit-font-smoothing: antialiased;
    -moz-osx-font-smoothing: grayscale;
}
```

2. The solution above wouldn't work for different values of `font-smoothing`. 

```css
/* non prefixed */
font-smooth: auto;
font-smooth: never;
font-smooth: always;
font-smooth: 2em; /* Length: no counter-parts */

/* webkit */
-webkit-font-smoothing: none;
-webkit-font-smoothing: antialiased;
-webkit-font-smoothing: subpixel-antialiased;

/* mozilla */
-moz-osx-font-smoothing: auto;
-moz-osx-font-smoothing: inherit;
-moz-osx-font-smoothing: unset;
-moz-osx-font-smoothing: grayscale;
```

The problem is that some values have no counter-parts on a different vendor, so it's hard to tell when one should correspond to the other. Also, their algorithms might be totally different. This is why I believe they shouldn't be part of atomic and maybe be part of a custom CSS specific to the project. These values may change how a text might look like if on dark or light backgrounds.

Note that there are many warnings and notes in the MDN page for font-smoothing, saying we shouldn't use it: https://developer.mozilla.org/en-US/docs/Web/CSS/font-smooth

I agree though, that not having them in Atomic can be a pain.

src-code:
Alright, this one has stagnated long enough.  I think I'm going to abandon it, since we can add these rules as custom rules as necessary for our unique use case.  And I agree that having vendor-specific variations isn't ideal.  

Really the only reason this became desirable was due to theming, where we'd want to do something akin to `<div class="Ff(theme1) Fsm(theme1) Fsmw(theme1) Fsmm(theme1)"> `  So as the font family varies, so might the font smoothing.  But I expect this is an advanced use case that most would not run into, and as I said before, since Atomizer allows custom rulesets, it's easy to add in after the fact.

cc/ @renatoi @thierryk

PR #213, opened 2015-07-13
Fixing findClassNames() string trim behavior

src-code:
Right now, if an Atomic classname is the first thing in a parsed file, `findClassNames()` improperly strips the first character of the classname because it assumes there should be a boundary character present.  

Switching to `trim()` seems like an easy fix, however I'm still not clear on why the regex is matching the boundary character in the first place.  It is probably better if we fix the regex to not return that char... @renatoi, any idea why this was necessary?

### DIFF
--- a/src/atomizer.js
+++ b/src/atomizer.js
@@ -86,7 +86,7 @@ Atomizer.prototype.findClassNames = function (src/*:string*/)/*:string[]*/ {
 
     while (match !== null) {
         // strip boundary character
-        className = match[0].substr(1);
+        className = match[0].trim();
 
         // assign to classNamesObj as key and give it a counter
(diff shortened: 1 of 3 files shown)


src-code:
Ah, didn't realize the boundary char could be something other than whitespace/newline... that complicates things a bit...

src-code:
Ok, I think a better way to go is to simply do a grouping on the part of the regex that actually represents the classname, rather than trying to match everything and trim later.

renatoi:
Yeah, the regex for the boundary character is something I've always wanted to remove it or make it better but didn't have time to think about a better solution.

This fix looks good to me, makes more sense :)

Issue #212, opened 2015-07-07
Atomizer.addRules() doesn't gracefully handle adding the same rule multiple times


src-code:
There are cases where you may want to package up custom rulesets and require them from multiple components in a project.  If each of those rulesets is added to Atomizer via `addRules()`, Atomizer will currently throw an error rather than gracefully recognizing that the rule definitions match, and therefore there's no conflict.  Would be better to have Atomizer do a deep-comparison of the rule definitions and only throw if the definitions differ.

src-code:
Addressed in PR #216

PR #214, opened 2015-07-13
Adding ability to specify rules via the command line

src-code:
```
$ atomizer 

  Usage: atomizer [options] [path]

  Options:

    -h, --help                          output usage information
    -V, --version                       output the version number
    -R, --recursive                     process all files recursively in the path
    -c, --config [file]                 source config file
    -r, --rules [file]                  custom rules file
    -o, --outfile [file]                destination config file
    -n, --namespace [namespace]         adds the given namespace to all generated Atomic CSS selectors
    -H, --helpersNamespace [namespace]  adds the given namespace to all helper selectors
    --rtl                               swaps `start` and `end` keyword replacements with `right` and `left`
    --ie                                adds old IE hacks to the output
    --verbose                           show additional log info (warnings)
```

Can specify multiple rule files by specifying multiple `-r` params:

```
atomizer -r ./rules1.js -r ./rules2.js my_file.html
```

FYI @renatoi

PR #215, opened 2015-07-13
Better detection of rule conflicts

src-code:
Addressing issue #212 
@dnukem @renatoi


src-code:
This is built off the branch used for PR #214, so we should probably merge that one first.

PR #216, opened 2015-07-13
Doing deep comparison of rules to better detect conflicting defintions

src-code:
Addressing issue #212.

fyi @renatoi @dnewcome

### DIFF
--- a/src/atomizer.js
+++ b/src/atomizer.js
@@ -35,24 +35,23 @@ function Atomizer(options/*:AtomizerOptions*/, rules/*:AtomizerRules*/) {
  */
 Atomizer.prototype.addRules = function(rules/*:AtomizerRules*/)/*:void*/ {
     rules.forEach(function (rule) {
-        if (
-            (rule.type === 'pattern'
-                && this.rulesMap.hasOwnProperty(rule.matcher)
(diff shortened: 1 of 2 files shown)


src/atomizer.js
```diff
@@ -36,10 +36,14 @@ function Atomizer(options/*:AtomizerOptions*/, rules/*:AtomizerRules*/) {
 Atomizer.prototype.addRules = function(rules/*:AtomizerRules*/)/*:void*/ {
     rules.forEach(function (rule) {
         if (
-            (rule.type === 'pattern' && this.rulesMap.hasOwnProperty(rule.matcher)) ||
-            (rule.type === 'helper' && this.helpersMap.hasOwnProperty(rule.matcher))
+            (rule.type === 'pattern'
+                && this.rulesMap.hasOwnProperty(rule.matcher)
+                && !_.isEqual(this.rules[this.rulesMap[rule.matcher]], rule)) ||
```
dnewcome: maybe we could add a getter on the prototype for `this.rules[this.rulesMap[rule.matcher]]` since it's operating on its own data and it would be self-documenting.

src-code: @dnewcome Are you suggesting something along the lines of `this.rules.get(rule.matcher)`?

PR #217, opened 2015-07-14
add SVG rules

3den:
https://github.com/yahoo/atomizer/issues/209#issuecomment-121396276 @src-code @renatoi @thierryk

### DIFF
--- a/src/rules.js
+++ b/src/rules.js
@@ -2697,7 +2697,7 @@ module.exports = [
         "type": "pattern",
         "name": "Fill (SVG)",
         "matcher": "Fill",
-        "allowParamToValue": true,
+        "allowParamToValue": false,
         "styles": {
             "fill": "$0"
         },
@@ -2707,7 +2707,7 @@ module.exports = [
         "type": "pattern",
         "name": "Stroke (SVG)",
         "matcher": "Stk",
-        "allowParamToValue": true,
+        "allowParamToValue": false,
         "styles": {
             "stroke": "$0"
         },
--- a/src/rules.js
+++ b/src/rules.js
@@ -2687,5 +2687,72 @@ module.exports = [
         "arguments": [{
             "a": "auto"
         }]
+    },
+    /**
+    ==================================================================
+    SVG
+    ==================================================================
+    */
+    {
+        "type": "pattern",
+        "name": "fill",
+        "matcher": "Fill",
+        "allowParamToValue": true,
+        "styles": {
+            "fill": "$0"
+        },
+        "arguments": [colors]
+    },
+    {
+        "type": "pattern",
+        "name": "stroke",
+        "matcher": "Stk",
+        "allowParamToValue": true,
+        "styles": {
+            "stroke": "$0"
+        },
+        "arguments": [colors]
+    },
+    {
+        "type": "pattern",
+        "name": "stroke",
+        "matcher": "Stkw",
+        "allowParamToValue": true,
+        "styles": {
+            "stroke": "$0"
+        },
+        "arguments": [{
+            "i": "inherit"
+        }]
+    },
+    {
+        "type": "pattern",
+        "name": "stroke-linecap",
+        "matcher": "Stklc",
+        "allowParamToValue": true,
+        "styles": {
+            "stroke-linecap": "$0"
+        },
+        "arguments": [{
+            "i": "inherit",
+            "b": "butt",
+            "r": "round",
+            "s": "square"
+        }]
+    },
+    {
+        "type": "pattern",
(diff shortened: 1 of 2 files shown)


src/rules.js
```diff
@@ -2687,5 +2687,72 @@ module.exports = [
...
         }]
+    },
+    /**
+    ==================================================================
+    SVG
+    ==================================================================
+    */
+    {
+        "type": "pattern",
+        "name": "fill",
```
src-code: The 'name' field is used when we generate the reference docs, so it needs to be formatted.  I'd go with something like "Fill (SVG)", etc for these

src/rules.js
```diff
@@ -2687,5 +2687,72 @@ module.exports = [
...
+        },
+        "arguments": [colors]
+    },
+    {
+        "type": "pattern",
+        "name": "stroke",
+        "matcher": "Stkw",
+        "allowParamToValue": true,
+        "styles": {
+            "stroke": "$0"
```
src-code: Should be `stroke-width`

src/rules.js
```diff
@@ -2687,5 +2687,72 @@ module.exports = [
...
+        "matcher": "Stk",
+        "allowParamToValue": true,
+        "styles": {
+            "stroke": "$0"
+        },
+        "arguments": [colors]
+    },
+    {
+        "type": "pattern",
+        "name": "stroke",
```
src-code: Should be "Stroke width (SVG)"

src/rules.js
```diff
@@ -2687,5 +2687,72 @@ module.exports = [
...
+            "b": "butt",
+            "r": "round",
+            "s": "square"
+        }]
+    },
+    {
+        "type": "pattern",
+        "name": "stroke-linejoin",
+        "matcher": "Stklj",
+        "allowParamToValue": true,
```
src-code: Is this accurate?  I don't think it allows arbitrary values... https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/stroke-linejoin

src/rules.js
```diff
@@ -2687,5 +2687,72 @@ module.exports = [
...
+        "styles": {
+            "fill": "$0"
+        },
+        "arguments": [colors]
+    },
+    {
+        "type": "pattern",
+        "name": "stroke",
+        "matcher": "Stk",
+        "allowParamToValue": true,
```
src-code: Is this one limited only to the values in `colors`?  If so, I think `allowParamToValue` should be false

src/rules.js
```diff
@@ -2687,5 +2687,72 @@ module.exports = [
...
+        },
+        "arguments": [{
+            "i": "inherit"
+        }]
+    },
+    {
+        "type": "pattern",
+        "name": "stroke-linecap",
+        "matcher": "Stklc",
+        "allowParamToValue": true,
```
src-code: I think this should be `false`: https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/stroke-linecap

src/rules.js
3den: i think stroke have have more than just the color, seems like you can specify all stroke props with this rule

src-code: https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/stroke

It seems to only accept a "paint" value, same as `fill`?

src-code:
So following up, `stroke` and `fill` take "paint" values, which could be a color or one of a few keywords or references to other defined gradients, etc:

http://www.w3.org/TR/SVG/painting.html#SpecifyingPaint

If we allow param-to-value as we currently do, then I guess we'd be covered on most of these (the `funciri` might be an issue though)  Or we can list out these keywords as valid arguments, and then disable `allowParamToValue` to restrict people from passing invalid values.

3den:
@src-code can we pass custom variables is allowParamToValue is set to false?

src-code:
@3den Yes - `allowParamToValue` simply controls whether you can pass an a value such as `1em`, `100px`, etc.

3den:
Got it, i will fix that

3den:
@src-code fixed

src-code:
@3den So if we disable `allowParamToValue` on those, we won't be able to pass valid Paint keywords like `none` and `currentColor`, I believe, so we'd have to add those explicitly to the `arguments` array on those rules.

3den:
@src-code i think you already have those in colors (https://github.com/yahoo/atomizer/blob/master/src/colors.js)

src-code:
@3den Oooh, I didn't see `currentColor` in there!  But's missing `none`, so we'd have to add that one.

src-code:
@3den :+1:

PR #219, opened 2015-07-17
Add fallback for IE background-color with rgba declarations

etast:
Add fallback for IE background-color with rgba declarations.

### DIFF
--- a/src/atomizer.js
+++ b/src/atomizer.js
@@ -321,6 +321,12 @@ Atomizer.prototype.parseConfig = function (config/*:AtomizerConfig*/, options/*:
                         if (prop === 'opacity') {
                             treeo.declarations.filter = 'alpha(opacity=' + parseFloat(value, 10) * 100 + ')';
                         }
+                        /* istanbul ignore else  */
+                        if (prop === 'background-color' && value && value.match(/^rgba/)) {
+                            treeo.declarations['background-color'] = 'rgb(' +
+                            value.match(/\d+/g).splice(0,3).join(',') +
+                            '); background-color: ' + value;
+                        }
                     }
                     if (value !== null) {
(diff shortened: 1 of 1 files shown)


src-code:
This change looks for `rgba()` declarations in custom values and modifies the output to have an initial `rgb()` solid-color declaration alongside it, eg:

```css
.Bgc\(steve\) {
  background-color: rgba(45,45,45,0.3);
}
becomes...
```css
.Bgc\(steve\) {
  background-color: rgb(45,45,45); 
  background-color: rgba(45,45,45,0.3);
}
```
This seems reasonable to me, though I do still wonder if stuff like this is better handled by a post-processor rather than in Atomizer.

src-code:
Part of the issue here is that the fallback is opinionated - rather than using IE filters to achieve opacity, it assumes you want a solid color.  Maybe not a bad opinion, but still seems like something that shouldn't be decided by Atomizer

fyi @renatoi

renatoi:
I agree. Even though it's worth considering adding it, I'd prefer using better tools that can help achieve fallbacks such as [PostCSS Color RGBA Fallback](https://github.com/postcss/postcss-color-rgba-fallback).

This PR is handling `background-color`, but there are also other properties such as `color`, `border`, `outline`, etc.

Consider using PostCSS for this as it addresses the problem better.

src-code:
@renatoi Good point, I had forgotten this was narrowly targeted to background-color... that'd be an issue too.  

I'm going to close this one out in favor of adding the above-referenced postcss plugin to the project's build tasks.  Seems we should add some documentation to explain what then the `--ie` flag really gets you - eg, how do we decide what belongs in Atomizer vs what is up to you via a post-processor.

Issue #221, opened 2015-07-20
Atomizer doesn't handle optional params well


src-code:
Some styles have optional values, for instance `border-spacing` can have one or two values:

`border-spacing: 1em`
`border-spacing: inherit`
`border-spacing: 1em 33%`

Currently in Atomizer if a rule specifies two arguments, and you don't pass the 2nd argument, then the regex placeholder is passed through:

```css
.Bdsp\(1em\) {
  border-spacing: 1em $1;
}
.Bdsp\(i\) {
  border-spacing: inherit $1;
}
.Bdsp\(1em\,33\%\) {
  border-spacing: 1em 33%;
}
```

Seems that Atomizer should strip placeholders if they're not found.

@renatoi @thierryk

src-code:
In v3.1.1

PR #222, opened 2015-07-20
Clean up declaration placeholders for optional params

src-code:
Addressing issue #221.  

I considered a couple of different approaches to solve this.  One was to loop on placeholders rather than supplied params so we can more easily replace empty params, and detect when a classname fails to provide the expected number of params.  Even go a step further and have type checking... But in the end, I decided to keep this simple and retain the hands-off approach Atomizer currently employs, trusting the developer to pass the correct params, while simply cleaning up any stray placeholders that may not be used due to their optional nature.  That way we can at least produce valid CSS if valid params are supplied in the classname, whereas without this fix we can't.

If we think it's valuable to have stricter type checking in the future, I think we'll want to tackle that as a new issue.  But my gut tells me that's probably overkill.

@renatoi

### DIFF
--- a/src/atomizer.js
+++ b/src/atomizer.js
@@ -356,7 +356,12 @@ Atomizer.prototype.parseConfig = function (config/*:AtomizerConfig*/, options/*:
                         treeo.declarations = null;
                     }
                 });
+                // If any of the arguments in the declaration weren't replaced, then we need to clean them up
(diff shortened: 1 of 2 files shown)


src-code:
@dnewcome

dnewcome:
agree to keep it simple. If we want type checking we probably want a different approach.

Talked to @src-code about this offline, LGTM

Issue #218, opened 2015-07-17
Add support for border-spacing


nspector:
Was just looking at the reference (http://acss.io/reference) and noticed there was no support for border-spacing. Would be great if it could be added! 

Thanks!

src-code:
Should be easy to add, I'll take care of this one soon.

src-code:
Blocked by issue #221

src-code:
In v3.1.1

PR #223, opened 2015-07-21
Border spacing

src-code:
Introduces support for `border-spacing`.  Depends on PR #222, since it has valid syntax with either one or two values in the declaration.

Addresses issue #218 

@renatoi @thierryk @nspector


src-code:
@dnewcome

dnewcome:
LGTM

PR #235, opened 2015-07-30
Add parsing support for class names following closing brace

snyamathi:
@src-code I'm not sure if this is a valid use-case because (in theory) you could always replace the dust syntax in the example with conditional Atomic css.

### DIFF
--- a/tests/atomizer.js
+++ b/tests/atomizer.js
@@ -42,6 +42,12 @@ describe('Atomizer()', function () {
             var expected = ['Pos(r)', 'Ov(h)', 'H(0)'];
             expect(result).to.deep.equal(expected);
         });
+        it('properly finds Atomic classnames inside Dust template conditionals', function () {
+            var atomizer = new Atomizer();
+            var result = atomizer.findClassNames('<div class="Pos(r) Ov(h) H(0) {?foo}D(n){/foo}"></div>');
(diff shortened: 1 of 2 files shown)


src-code:
To clarify: The purpose of this change is to allow the parser to correctly locate Atomic classes where the classname is preceded by a closing brace, such as inside a Dust template conditional.

tests/atomizer.js
```diff
@@ -42,6 +42,12 @@ describe('Atomizer()', function () {
             var expected = ['Pos(r)', 'Ov(h)', 'H(0)'];
             expect(result).to.deep.equal(expected);
         });
+        it('returns an array of valid atomic class names if the boundary character before the first found classname is a closing brace (Dust.js)', function () {
```
src-code: This seems to be a cut and paste error - the notion of "first found classname" doesn't apply here (in another test, that refers to a classname that's the first thing found in a file).  I'd change this to something like:

"properly finds Atomic classnames inside Dust template conditionals"

Or if there are other use cases besides Dust templates that could result in curly braces as boundary chars, you could adjust accordingly.

snyamathi:
@src-code thanks, I've updated the test label.

src-code:
:+1:

src-code:
This'll land in v3.2.0

PR #243, opened 2015-10-26
Support background position center bottom

roderickhsiao:
@src-code @renatoi 

Seems like we are missing one attribute `background-position: center bottom`


src-code:
LGTM.  Also seems we're missing `end_c`, so might be worth adding that along with this pull.

src-code:
Thanks @roderickhsiao!

PR #248, opened 2015-11-17
Updating import of XRegExp

src-code:
The old method of importing XRegExp seems to be breaking on the client now with the 3.0.0 release, even though the docs claim it's still valid.  I'll just update to use the simplified `require()`, since it fixes the issue.


renatoi:
:+1:

roderickhsiao:
:+1:

PR #251, opened 2015-12-07
Add inline-flex to display

lukebrooker:
Will match `D(if)`.


roderickhsiao:
:+1: @renatoi @src-code how do you think?

whmountains:
:+1:

PR #253, opened 2016-01-04
Added atomizer logo to readme

RichardBray:
Let me know if you think the image should be smaller.


redonkulus:
I'll defer to @src-code on this but the size is quite large for the readme. Also there is a .DS_Store file that should not be committed.

RichardBray:
Good point, will sort out the DS file and logo size some time today (UK time).

src-code:
Really this logo is for Atomic CSS, not for Atomizer.  Though we don't really make the distinction very clear on the web site (acss.io), I suppose.  So I'm not sure we'd want to use the logo for Atomizer, but perhaps @renatoi or @thierryk think otherwise?

RichardBray:
Ah, that makes sense. Would it be possible to get the name of the person who made the logo? I would like to give them credit for it.

PR #257, opened 2016-03-24
Update grammar definitions to support CamelCase breakpoint names.

alexnj:
As the title says, this would allow breakpoint definitions to have CamelCase names.


src-code:
:+1: Docs don't specify valid case, so no changes needed there.

thierryk:
:+1:

GanbilegBor-minted:
👍

PR #258, opened 2016-04-04
Fixing grammar for classes with numbers in their prefix

src-code:
`Translate3d(...)` wasn't working property because numbers weren't being allowed in the classname prefix.


xvash:
:shipit:

PR #261, opened 2016-06-04
Move to new acss-io organization

okuryu:
The atomizer seems to moved from yahoo org to acss-io org.


yahoocla:
CLA is valid!

renatoi:
Thanks!

Issue #246, opened 2015-10-28
Uppercase hex values are silently ignored


src-code:
From @thierryk:

> Today, the class `Bgc(#0280AE.5)` produces this style: `background-color: #028;`. But I think it should fail as it uses uppercase letters.

I believe we disallowed uppercase letters in order to avoid a scenario where mixed case would lead to duplicate styles being created with slightly different selectors.  If we're going to continue disallowing uppercase in hex values, then it seems like they should at least throw a warning, if not break outright.  (Since the resulting CSS is not correct.)  Or perhaps we allow uppercase but throw a warning?  I might prefer that approach, actually...

Thoughts?  @renatoi @dnewcome

dnewcome:
Ideally I think we'd want to accept upper or lower case and normalize the generated CSS rule to one or the other.

That probably complicates things since we use the same code paths (I think) for other case-sensitive values. I think we should generate a warning at the very least giving the generated CSS. We could support mixed case in the future, or change the warning to an error and that's the end of it.

src-code:
The problem with normalizing it automatically is that we'd have to modify the markup as well, otherwise the classes won't match the selectors and the CSS will be broken.  So I'm inclined to support case-sensitive hex values, but throw a warning if they're detected, that way at least we're not introducing subtle bugs in the CSS, but we're also pushing developers to optimize their code further.

samlecuyer:
This is somewhat fixed by #263

thierryk:
Fixed by https://github.com/acss-io/atomizer/pull/263

PR #263, opened 2016-06-08
Validate hex strings containing uppercase digits, emit warning

samlecuyer:
Currently, the grammar does not support case insensitive hex.
So, `#fff` parses while `#FFF` does not.  This fixes that.
Unfortunately, because the HEX definition is nested inside of
a large `String.join()`, we can't just add a case insensitive flag.
Without refactoring the entire grammar, this seems like the most
expedient fix.

@src-code /cc @ajcrews

### DIFF
--- a/src/atomizer.js
+++ b/src/atomizer.js
@@ -236,7 +236,10 @@ Atomizer.prototype.parseConfig = function (config/*:AtomizerConfig*/, options/*:
                     value = Math.round(matchVal.numerator / matchVal.denominator * 100 * 10000) / 10000 + '%';
                 }
                 if (matchVal.hex) {
-                    if (matchVal.alpha) {
+                    if (matchVal.hex !== matchVal.hex.toLowerCase()) {
+                        console.warn('Warning: Only lowercase hex digits are accepted. No rules will be generated for `' + matchVal.input + '`');
+                        value = null;
+                    } else if (matchVal.alpha) {
                         rgb = utils.hexToRgb(matchVal.hex);
                         value = [
                             'rgba(',
--- a/src/lib/grammar.js
+++ b/src/lib/grammar.js
@@ -71,7 +71,7 @@ var GRAMMAR = {
     'PARAMS'        : '\\((?<params>[^)]*)\\)',
     'NUMBER'        : '-?[0-9]+(?:\.[0-9]+)?|\\.[0-9]+',
     'UNIT'          : '[a-zA-Z%]+',
-    'HEX'           : '#[0-9a-f]{3}(?:[0-9a-f]{3})?',
+    'HEX'           : '#[0-9a-fA-F]{3}(?:[0-9a-fA-F]{3})?',
     'ALPHA'         : '\\.\\d{1,2}',
     'IMPORTANT'     : '!',
     // https://regex101.com/r/mM2vT9/8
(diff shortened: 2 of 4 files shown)


yahoocla:

thierryk:
For what it is worth, I don't like this for 2 reasons:

   * it goes against most styles guides (including the one I wrote for Yahoo!)
   * it opens the door for redundancy, which goes against Atomic CSS philosophy

To me, it would make more sense to create a flag that would allow the use of one or the other, but not both.

samlecuyer:
If that's the case, I think that this should be a warning or an error.  As it is, it doesn't generate an error, a warning, or the expected output. The RegExps in `src/lib/utils#hexToRgb` are specifically case-insensitive, which may lead users to think that maybe the grammar is too.

Generally, I think silent partial failures are a bad thing.  Style guides should probably be enforced by a linter in a later step, or the fact that this is a feature should be documented.

arjansingh:
I agree with @samlecuyer. For transpilation libraries like this, any syntax errors should be at least warnings and probably noisy failures. If CAPS hex values are truly a syntax error, they should be treated accordingly.

thierryk:
> If CAPS hex values are truly a syntax error, they should be treated accordingly.

I think everybody agrees on this but this is not what this PR is about.

yahoocla:
CLA is valid!

ajcrews:
👍

samlecuyer:
I think that _is_ what this PR is about.  Atomizer occupies a somewhat precarious position where it has to generate output for only some of its input and decide on how to report anything it skips.

The main issue here is that atomizer fails to parse what are effectively valid CSS property values without any mention of this failure.  This is a pretty far cry from not mentioning that `.vp-wrapper` isn't getting any rules generated.  There's no real technical reason why "CAPS hex" can't be parsed, but because it simply _doesn't_ parse, no messages are generated.  

I think that ideally there should be a few separate steps:

 1) Parse as many valid css properties as possible
 2) Validate the properties according to some ruleset and prune
 3) Report what was pruned and why

This PR isn't a perfect fix, but I think that accepting `#01F2D3` as valid input is a good first start to not necessarily processing it for codegen.

thierryk:
> I think that accepting #01F2D3 as valid input is a good first start 

@samlecuyer I agree with the steps you listed, but I don't agree with the above statement which seems to go against the [original suggestion](https://github.com/acss-io/atomizer/issues/246#issuecomment-152272348)

I definitely think a _warning_ (or even an *error*) is a better approach than allowing for **both** uppercase and lowercase letters.

As a side note, we have the same problem with color keywords. I don't really expect people to write those keywords in uppercase letters (i.e. `C(RED)`) but as we all agree, failing silently is pretty bad...

samlecuyer:
I think we may have gotten a little far into the weeds on this PR, so let's recenter this discussion.

A few things we can all agree on:

 * There are certain valid css values that are not valid acss values. CAPS hex is in this.
  - These values should ideally generate a warning
 * Atomizer can only report validation errors on tokens it parses.

Given that, this PR adds CAPS hex to the grammar so that we can know about it.  Additionally, it validates that hex colors are lowercase and generates a warning instead of a value in that case.

I'm renaming this to reflect the changes and rebasing the commits as well.  Is this an acceptable contribution?

src-code:
@samlecuyer 👍 I think this is the right way to go!

thierryk:
Thanks @samlecuyer 
I got stuck on the original title of this PR...

Issue #260, opened 2016-06-03
ordering issues with atomic parser


modosc:
we had some styling issues with our site when using [atomic-plugin] (https://github.com/mdrohmann/atomic-plugin) but the site worked fine when using `atomizer` on the command line. i narrowed it down to the order in which files were being parsed - inside of `webpack` files were concatenated together based on the module require tree rather than the glob we were using on the command line.

example:

```
> var test1 = '<div class="D(n)"></div><div class="D(ib) D(n)">test</div>'
> var test2 = '<div class="D(ib) D(n)">test</div><div class="D(n)"></div>'

> console.log(atomizer.getCss(atomizer.getConfig(atomizer.findClassNames(test1), {})))
.D\(n\) {
  display: none;
}
.D\(ib\) {
  display: inline-block;
}

> console.log(atomizer.getCss(atomizer.getConfig(atomizer.findClassNames(test2), {})))
.D\(ib\) {
  display: inline-block;
}
.D\(n\) {
  display: none;
}
```

`test1` will be rendered with `display: inline-block`, `test2` with `display: none`.

we wound up calling `sort()` on the classes that we pass to `atomizer.getConfig` and that resolved the issue by making sure that `.D(n)` always had priority over `.D(i)`, `.D(ib)`, etc but that's just dumb luck that `none` sorts last. i think there should be some explicit ordering of these classes when they're generated so that the same files always render the same, regardless of the order of parsing.

modosc:
another example is with media queries - using this config:
```
  breakPoints: {
    sm: '@media screen and (max-width: 900px)'
  }
```
these two blocks display differently:
```
<div class="W(1/3) W(100%)--sm">
```

```
<div class=" W(100%)--sm W(1/3)">
```

keep in mind it's not just the ordering of the classes here - if a different file references one or other other class _before_ this one it will change the order of the generated css and thus the behavior. 

also the second example's css looks like this:
```
@media screen and (max-width: 900px) {
  .W\(100\%\)--sm {
    width: 100%;
  }
}
  .W\(1\/3\) {
    width: 33.3333%;
  }
```

(note the indenting after the `@media` section)

dnewcome:
[...]

PR #265, opened 2016-07-06
JSS: Processing main block before media query blocks

src-code:
Addressing issue #260.  Prioritizing the processing of the main block before attempting to process any media query blocks.  

@renatoi @thierryk


yahoocla:
CLA is valid!

src-code:
Test case based on https://github.com/modosc/atomizer-test-case added.

Issue #267, opened 2016-07-26
feature request: add css contain property


yurited:
Chrome 52 releases `contain: none | strict | content | [ size || layout || style || paint ]` letting developers limit the scope of the browser’s styles, layout and paint work.
https://developers.google.com/web/updates/2016/06/css-containment?hl=en
http://caniuse.com/#search=css%20containment

![hero](https://developers.google.com/web/updates/images/2016/06/containment.jpg)

PR #268, opened 2016-07-27
add rule for contain

yurited:
Chrome 52 releases `contain: none | strict | content | [ size || layout || style || paint ]` letting developers limit the scope of the browser’s styles, layout and paint work.
https://developers.google.com/web/updates/2016/06/css-containment?hl=en
http://caniuse.com/#search=css%20containment

hero

https://github.com/acss-io/atomizer/issues/267


src-code:
I'm generally ok with this although it's a rather early spec.  I've seen Firefox intends to implement it, but curious if any others will follow.

One note: Emmet, which we try to follow as closely as possible when choosing abbreviations, uses both `Cnt` and `Ct` for `content`.  They don't yet offer `contain`, so it'll be interesting to see what they adopt when they support this.  

http://docs.emmet.io/cheat-sheet/

src-code:
I like this better with `Ctn`!  👍

Issue #272, opened 2016-09-27
LineClamp causing child inline-block elements to wrap


src-code:
It was reported to me that there's a use case where `LineClamp()` is causing child elements styled as `inline-block` to wrap incorrectly due (presumably) to the use of `-webkit-box` on the clamped node.  For example:

```html
<div class="LineClamp(2,36px)">
<span>Let's clamp this long piece of text!</span>
<span class="D(ib)">This line should not wrap</span>
</div>

```

```css
div {
  border: 1px solid #000;
  line-height: 18px;
  width: 200px;
  height: 300px;
}

/**
 * Generated using Atomizer build v3.2.1. 
 * Copyright (c) 2015, Yahoo Inc. All rights reserved. **/

[class*=LineClamp] {
  display: -webkit-box;
  -webkit-box-orient: vertical;
  overflow: hidden;
}

a[class*=LineClamp] {
  display: inline-block;
  display : -webkit-box;  
  *display: inline;
  zoom: 1;
}

a[class*=LineClamp]:after {
  content: ".";
  font-size: 0;
  visibility: hidden;
  display: inline-block;
  overflow: hidden;
  height: 0;
  width: 0;
}

.LineClamp\(2\,36px\) {
  -webkit-line-clamp: 2;
  max-height: 36px;
}

.D\(ib\) {
  display: inline-block;
}
```

This will cause the second span to appear on a line below the first span, which causes it to be hidden thanks to the line clamp.

Test case: https://jsfiddle.net/ratLrm0L/1/

One solution seems to be to make the clamped node `display: inline-block`, but not sure this is the correct solution.  @thierryk, any thoughts?

src-code:
fyi @uribe @willthai

fabianuribe:
Thanks for looking into it @src-code  ;)

thierryk:
@src-code, @fabianuribe I'm checking the test case in FF 48 and I do not see any issue there. Could it be a bug that was fixed in later releases?
Could you guys check in the latest version (49 I believe) and let me know if it works for you or not?
Thanks!

src-code:
@thierryk Doesn't work in FF 49.  Didn't test in earlier FF, but I did note that `-webkit-box` was supported as of FF 45, so I assumed it was the same for all versions older than that.

fabianuribe:
Correct it just started breaking with 49.0.1, from the release notes:

```
The following display values are translated:
-webkit-box to -moz-box
-webkit-flex to flex
[...]

PR #273, opened 2016-09-28
Adding support for @supports, fixing LineClamp for FF49+

src-code:
Adding support for `@supports` to JSS, so that we can fix `LineClamp()` in FF 49+ (see issue #272).

fyi @fabianuribe @thierryk @renatoi


fabianuribe:
👍

src-code:
Updated to use `block` instead of `inline-block`

renatoi:
:shipit:

PR #274, opened 2016-09-30
feat(grammar): accept hyphenated-prop name

sarbbottam:
if custom `rules.js` specifies `"font-size"` matcher as `f-z`
and `f-z(16px)` is used, following css should be generated
```css
.f-z\(16px\) {
    font-size: 16px;
}
```
cc/ @renatoi

### DIFF
--- a/src/lib/grammar.js
+++ b/src/lib/grammar.js
@@ -64,7 +64,7 @@ var GRAMMAR = {
     'PARENT'        : '[a-zA-Z][-_a-zA-Z0-9]+?',
     'PARENT_SEP'    : '[>_+]',
     // all characters allowed to be a prop
-    'PROP'          : '[A-Za-z0-9]+',
+    'PROP'          : '[-A-Za-z0-9]+',
     // all character allowed to be in values
     'VALUES'        : '[-_,.#$/%0-9a-zA-Z]+',
     'FRACTION'      : '(?<numerator>[0-9]+)\\/(?<denominator>[1-9](?:[0-9]+)?)',
--- a/tests/atomizer.js
+++ b/tests/atomizer.js
@@ -42,6 +42,21 @@ describe('Atomizer()', function () {
             var expected = ['Pos(r)', 'Ov(h)', 'H(0)'];
             expect(result).to.deep.equal(expected);
         });
+        it('properly resolves custom classnames with hyphen, in custom rules.js', function () {
+            var rules = [{
+                type: 'pattern',
+                name: 'Font Size',
+                matcher: 'f-z',
+                styles: {
+                    'font-size': '$0'
+                }
+            }];
+            var atomizer = new Atomizer();
+            atomizer.addRules(rules);
+            var result = atomizer.findClassNames('<div class="Fz(RWD-fontSize) f-z(RWD-fontSize)"></div>');
+            var expected = ['Fz(RWD-fontSize)', 'f-z(RWD-fontSize)'];
+            expect(result).to.deep.equal(expected);
+        });
         it('properly finds Atomic classnames inside Dust template conditionals', function () {
             var atomizer = new Atomizer();
             var result = atomizer.findClassNames('<div class="Pos(r) Ov(h) H(0) {?foo}D(n){/foo}"></div>');


renatoi:
👍  I don't see any problems with this.

thierryk:
Lower case hyphenated-prop name? What is the value of doing this? What are the advantages?

sarbbottam:
Personal preference for naming `css` class. I beleive a user should have the flexibility to customize a tool. [I always wanted to try atomizer, but stayed away due to `PascalCase`](https://twitter.com/sarbbottam/status/720054008795176960).

I hope this would get merged, unless it causing a regression.

pankajparashar:
I'm not familiar with the codebase, but @renatoi could confirm if would this clash with custom config parameters that include `-` in the name?

For example,
```
<div class="Fz(RWD-fontSize)">Responsive font-size</div>
```

Sourced the above example from [acss.io](http://acss.io/)

thierryk:
@sarbbottam this is why I asked. Because the syntax was originally chosen to answer many challenges. To be honest, we were more worried about extending the abilities of the tool than pleasing dev's eyes.

> atomizer is perfect, but the grammar is not intuitive IMHO.

> I have been staying away just for the naming convention but was always sold out to the idea.

It looks like @renatoi does not agree with me on this one but to me, that's not a valid reason to open Pandora's box as this change may not be as trivial as it seems. @renatoi and @src-code should weigh in&mdash;is that something we could regret down the road because of new CSS feature, new syntax, or things we did not test against yet?

Finally, I've always been a big opponent of anything that could generate *bloat*. For example, I was against allowing lowercase and uppercase for color values. Because such approach leads to rules like this one:

``` css
.Fz\(16px\), 
.f-z\(16px\) {
    font-size: 16px;
}
```

I don't like making changes **to the grammar** for reasons like this. If people are that anal about it, I think they should fork the project and make changes that fit their sensibilities. Because there will be somebody to tell us that `fz` is much better than `f-z` because the `-` clashes with their own naming convention for their classic/semantic classes. 

At the end of the day, we could get this:

``` css
.Fz\(16px\),
.F-z\(16px\), 
.fz\(16px\), 
.f-z\(16px\) {
    font-size: 16px;
}
```

We have style guides to make sure devs obey basic rules (do this, don't do this, etc.). So why not looking at Atomizer as a tool that enforces a specific style guide for devs' own good?

renatoi:
@thierryk the rules file allow them to change the matcher of the class to whatever they want already, the only issues is that it currently does not accept the dash.

It's important to note that this is not changing or allowing ACSS classes to have variations, but replacing them completely. So situations like this would never happen:

```css
.Fz\(16px\), 
.f-z\(16px\) {
    font-size: 16px;
}
```

This is because Atomizer still follows [rules for each and every property](https://github.com/acss-io/atomizer/blob/master/src/rules.js#L1362). If the author wants to change the matcher of the `font-size` to the following they can:

```css
.font-size\(16px\) {
    font-size: 16px;
}
```

The issues I can see is that if the author does this and decides to create their own dictionary for acss classes, then they won't have much benefit from the tools that @pankajparashar created or even the reference in acss.io, for example. We can't also guarantee that the parse would work properly since the tests won't be covering their own dictionary.  But I believe this would be a known consequence and maybe help the author realize that the original syntax might appeal to them.

@pankajparashar Regarding your concern, I don't remember exactly (and I'm bit lazy to look at the code) but I believe this won't affect because it's the props matcher and not the values matcher. In any case, I believe it's a good idea for @sarbbottam to add a test [here ](https://github.com/acss-io/atomizer/blob/master/tests/atomizer.js#L31) to make sure that this won't break.

renatoi:
@thierryk just to clarify, this is not to conflict with any existing acss projects, but for people who wants to change the syntax for their own project and won't be using the original acss syntax at all.

thierryk:
@renatoi thanks for clarifying this to me as I obviously missed that part about redundancy but I still think it's better to avoid different syntax "flavors"&mdash;*even if these differences were meant to be found in unrelated projects*.

src-code:
I don't have a problem with this change, though I don't want or expect to see hyphens introduced into our default ruleset.  I agree with @renatoi  though that @sarbbottam should introduce tests to ensure this works and doesn't inadvertently interfere with matching other string patterns.

sarbbottam:
@renatoi @src-code please have a look. Thanks!

src-code:
Thanks @sarbbottam - @renatoi, @thierryk  or @pankajparashar, any further thoughts on this one?

thierryk:
For the record, I think this is ridiculous...

pankajparashar:
I would give up a **thumbs up** to this proposal, mainly for the following reasons,

1. This change in no way affects the way _Atomizer_ works out of the box either for new users, or the ones already using it.
2. This change is practically useless, unless someone introduces a new rule. Also, you need to know your way around the codebase, to make this work, which means you already know what you're doing at your own risk.
3. However, what this change does offer, is for experienced devs like @sarbbottam, a chance to use Atomizer the way they want to, thus offering extensibility.

In hindsight, extensibility is a good thing for the overall popularity of the framework.

thierryk:
One of the main advantages of Atomic CSS is the fact that teams/third-parties/etc. can exchange snippets of markup and get design patterns out-of-the-box (i.e. a carousel, etc.) **for free**.
Thinking this change is a good thing shows we are abandoning the idea that that is a great feature.
I want to remind everybody that&mdash;at Yahoo!&mdash;we saw Atomic CSS as a mean to accept CSS from third-party vendors without risk of breakage. 
This is what we'll be losing for allowing people to change the syntax because it's "too hard on their eyes".

sarbbottam:
@thierryk with all due respect to your css expertise I don't agree with the following.
>One of the main advantages of Atomic CSS is the fact that teams/third-parties/etc. can exchange snippets of markup and get design patterns out-of-the-box (i.e. a carousel, etc.) for free.

This statement assumes that the team/third-parties are also following the Atomic.css naming convention. For teams I can understand the organizational guidance. But what guarantees that third-parties are following the same convention?

Speaking about organization, they will use only one naming pattern, either `Atomic.css naming pattern` or `customized naming pattern`. Similar to shared `jshint`, `eslint` config, at least Yahoo had that.

I find the repeated appearance *dev's/peoples eyes* harsh and insulting.
> we were more worried about extending the abilities of the tool than pleasing *dev's* *eyes*
>allowing *people* to change the syntax because it's "too hard on their *eyes*"

It's not pleasing but asking people not to change their reflex, subconscious.

I am sorry to say, your arguments echoes `jslint`s ideology, _you must use my code formatting rules_ (`tabs` instead of `spaces` for indentation, etc); no customization; use it or leave it.
I feel this is where `eslint`/`jshint` own the race. 

This PR has no intention to change the core principles of Atomic css but to customize the way one is comfortable and productive. 

>We have style guides to make sure devs obey basic rules (do this, don't do this, etc.). So why not looking at Atomizer as a tool that enforces a specific style guide for devs' own good?

Yep, not every dev will be writing a a custom rules.js but an organization. If my organization enforces Atomic.css rules, I am more than happy to get aligned. But Why on my personal project?

I hope I am able to express myself and you would be open to accept this PR.

thierryk:
> This statement assumes that the team/third-parties are also following the Atomic.css naming convention. For teams I can understand the organizational guidance. But what guarantees that third-parties are following the same convention? 

Stencil at Yahoo! was enforcing a unique syntax, the same thing Atomizer should do. Today, if you go to sports.yahoo.com or finance.yahoo.com and copy/paste snippets of markup into [@pankajparashar's tool](https://pankajparashar.com/atomizer-web/) you'd get the exact same result. Allowing various flavors of the syntax goes against the idea of sharing snippets across the board.

> It's not pleasing but asking people not to change their reflex, subconscious.

I'm sorry if you felt offended but that's the way I see it. I'm honestly really tired of seeing devs focusing on details that relate to their own comfort rather than seeing the big picture. Please take a look at [this issue](https://github.com/necolas/normalize.css/pull/588) on normalize.css to better understand my frustration.

> This PR has no intention to change the core principles of Atomic css but to customize the way one is comfortable and productive.

`Fz` makes you uncomfortable and less productive? ok...

> If my organization enforces Atomic.css rules, I am more than happy to get aligned. But Why on my personal project?

Because if you create an awesome component, people won't be able to copy/paste your code to get the same result. And on your end, you won't be able to use [@pankajparashar's online tool](https://pankajparashar.com/atomizer-web/). You may think you have no need for this but you may find it useful down the road (it's very handy). 

> I hope I am able to express myself and you would be open to accept this PR.

It seems everybody but me is onboard with the idea so l think this PR should go through. 

I just hope the next PR we'll get won't be for *allowing uppercase hexadecimal color code*.

sarbbottam:
Thanks @thierryk.

_I might embrace the Atomic.css grammar if I find more trouble than values with the custom class names. Allow me to experience it_ 😄

thierryk:
> I might embrace the Atomic.css grammar if I find more trouble than values with the custom class names.

You are missing the point, this should not be about you but about the "*ACSS community*". People should be able to re-use any markup they find in the wild&mdash;be it on sports.yahoo.com or sarbbottam.com.

sarbbottam:
> People should be able to re-use any markup they find in the wild—be it on sports.yahoo.com or sarbbottam.com.

Only if sports.yahoo.com and sarbbottam.com has followed Atomic.css convention. 
But there is no guarantee that sports.yahoo.com and sarbbottam.com have used Atomic.css.

Sorry to differ, I don't think nowadays markup is not treated as it is own, but as a component, that encapsulates markup, style and script. Then again web technologies/utilities changes, so often, that I might be wrong in near future.

pankajparashar:
@sarbbottam Let's assume if this PR is accepted, 

How do you plan to use it? Simply updating the regex wouldn't help you here. 

At some point in future, you would want to update the grammar with your syntax rules, which we cannot accept here, since, it would create a conflict with our rules.

or If you are planning to maintain a separate fork for this project, then this PR kinda makes it redundant, isn't it?

sarbbottam:
>How do you plan to use it? Simply updating the regex wouldn't help you here.

I would maintain my `rules.js`, which atomizer respects at this point in time.

>At some point in future, you would want to update the grammar with your syntax rules

No, I want to use it as is, but customize the rules. I could do `fz(16px)` instead of `Fz(16px)` but not `f-z(16px)`, this PR is facilitating it.

> If you are planning to maintain a separate fork for this project, then this PR kinda makes it redundant, isn't it

No, I don't want to create fragmentation. I would only maintain my `rule.js` similar to custom `.jshint`, `.eslint` config file.

thierryk:
@renatoi @pankajparashar @src-code 

FYI: I don't want to get involved with threads like this in the future so I decided to *unwatch* this repo and *unsubscribe* as well.

src-code:
@thierryk I'm sorry to hear that, and sorry that this thread is causing you agitation!  Can't blame you for wanting to stay out of it.

src-code:
Ok, let's recap and put this to rest.  Already,

   * Users of Atomizer already have the ability to create custom **configuration**, which must be bundled with a project in order to make the project's markup portable.
   * Users of Atomizer already have the ability to create custom **rulesets** and **helpers** (as supported by the `-r` param to Atomizer), which also would need to be bundled with a project in order to make the project's markup portable.
   * Any developer who wants to consume another project's Atomizer-enhanced markup would already always need to consume any configuration and rulesets provided in that project.
   * This pull request changes none of these facts.
   * This pull request changes the characters that are considered valid for classname prefixes by adding a single character to the regex.

For these reasons I see no valid argument against this pull request, so I'm going to merge it.  Adding a hyphen does nothing to make Atomizer markup less portable than it already is. 

That said, Thierry is right in one regard - users of Atomizer should be strongly discouraged from customizing the default class prefix matchers if they do intend to share markup with OSS projects, or consume markup from other OSS projects.  Why?  Because they'd only be bloating their markup (and/or others') by creating multiple class definitions that map to the same style (eg, `.f-z(16px)` and `.Fz(16px)` both mapping to `font-size: 16px` in the resulting spreadsheet.)  

In short, **custom rulesets should only be used to create new, previously unsupported rules and helpers, and not used to "fork" the default set of rules.**  This should be a best practice documented somewhere on acss.io, so developers understand the problems involved in forking the default ruleset.

sarbbottam:
Thanks @src-code.

sarbbottam:
This PR was a mistake IMHO, better late than never for me to realize it.

I was too biased to realize that forking the grammar is actually creating the bloat which Atomic.CSS wanted to remove. @thierryk was absolutely correct that the grammar must not be altered. My bad, I didn't realize. Recently, I have been working on a presentation on "Evolution of Styling" and I believe Atomic.CSS is the answer to most of the CSS problem we have today, by not writing CSS, rather generating it. 

While considering the internal and external component ecosystem, built on Atomic.CSS principle, I realized my PR was a mistake. If all the component starts using their own grammar it would create unnecessary bloat, in the consuming application.

Sorry for all the churn. I hope you would excuse me.

PR #276, opened 2016-11-17
Add `appearance` to Atomizer.

soluml:


soluml:
Is this good to go? @src-code

src-code:
@soluml Following Emmet syntax, the matcher should be `Ap`, otherwise I'm fine with this.

soluml:
I'll go ahead and update the matcher.

src-code:
👍

PR #282, opened 2016-12-29
Add placeholder pseudo element

soluml:
Add a [placeholder pseudo element](https://developer.mozilla.org/en-US/docs/Web/CSS/::placeholder) to Atomizer.

### DIFF
--- a/src/lib/grammar.js
+++ b/src/lib/grammar.js
@@ -43,7 +43,8 @@ var PSEUDO_ELEMENTS = {
     '::before':         '::b',
     '::after':          '::a',
     '::first-letter':   '::fl',
-    '::first-line':     '::fli'
+    '::first-line':     '::fli',
+    '::placeholder':    '::ph'
 };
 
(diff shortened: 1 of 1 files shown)


soluml:
[Chrome looks to be adding it in 57](https://bugs.chromium.org/p/chromium/issues/detail?id=623345) / Opera in 43. Also looks to be in the TP of Safari ([caniuse](http://caniuse.com/#feat=css-placeholder)).  It's at Working Draft in the [spec](https://drafts.csswg.org/css-pseudo-4/#placeholder-pseudo) and Autoprefixer does a great job of handling the prefixed variants.

soluml:
Everything good @redonkulus?

redonkulus:
@soluml I will wait for @src-code to check this before merging. He is on vacation but will be back next week.

src-code:
Sounds good.  We considered this one back during #236 but didn't end up adding it, I guess due to poor browser support.  With Safari and Chrome adopting it soon, it seems like a good time to support it.

PR #283, opened 2017-02-10
Defect fixed where using Px, Mx etc. with a breakpoint configuration …

johnnorris:
We've added two tests to demonstrate this issue, but ultimately it's just that only one declaration was set (the last one) for breakpoint variables when using `Px`, `Py`, `Mx`, `My` etc.

E.g.
```
'Py($gutter)': {
    default: '10px',
    sm: '16px',
    md: '20px',
    lg: '24px'
}

// Produces:

@media screen and (min-width:700px) {,
  .Py\\(\\$gutter\\) {,
    padding-bottom: 16px;,
  },
},
```

We've added a very simple one line fix for this.

### DIFF
--- a/src/atomizer.js
+++ b/src/atomizer.js
@@ -346,7 +346,7 @@ Atomizer.prototype.parseConfig = function (config/*:AtomizerConfig*/, options/*:
                                 if (!config.hasOwnProperty('breakPoints') || !config.breakPoints.hasOwnProperty(bp)) {
                                     return;
                                 }
-                                treeo.declarations[config.breakPoints[bp]] = {};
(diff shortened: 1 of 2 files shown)


src-code:
Hi @johnnorris, thanks for the PR.  I had forgotten this syntax was supposed to be possible, I don't think it's documented anywhere.  ([Would be good to document it...](https://github.com/acss-io/acss-site))  

Cc'ing @renatoi to take a look, as I think he wrote this code originally.

renatoi:
Yes, I think I've missed this originally. Thanks for catching it and updating it. :shipit:

PR #286, opened 2017-03-20
Media query blocks in output CSS aren't ordered

johnnorris:
# The problem

Given an input of:

```
.element {
    @media screen and (min-width: 1000px) {
        color: white;
    }
}

.element {
    @media screen and (min-width: 600px) {
        width: 200px;
    }
}

.element {
    @media screen and (min-width: 1000px) {
        width: 500px;
    }
}
```

You currently get an output of:

```
@media screen and (min-width: 1000px) {
    .element {
        color: white;
        width: 500px;
    }
}

@media screen and (min-width: 600px) {
    .element {
        width: 200px;
    }
}
```

This we think is because the first 1000px entry was before the first 600px entry, and they're grouped up.

At a screen width of more than 1000px now, your element has a width of 200px instead of the expected 500px, because of the order of the media query blocks.

# What is expected?

We expect that the smallest media query blocks are higher in the CSS file than the larger ones so that the larger ones override the smaller ones.

The names of the breakpoints are arbitrary enough that it's difficult to sort by them. You could sort by the actual breakpoint values, but that would be quite complex, so as a compromise we decided to order them in the same order as the breakpoints you configure.

### DIFF
--- a/src/atomizer.js
+++ b/src/atomizer.js
@@ -534,7 +534,9 @@ Atomizer.prototype.getCss = function (config/*:AtomizerConfig*/, options/*:CSSOp
     });
 
     // convert JSS to CSS
-    content = options.banner + JSS.jssToCss(jss);
+    content = options.banner + JSS.jssToCss(jss, {
+      breakPoints: breakPoints
+    });
 
     // fix the comma problem in Absurd
     content = Atomizer.replaceConstants(content, options.rtl);
--- a/src/atomizer.js
+++ b/src/atomizer.js
@@ -346,7 +346,7 @@ Atomizer.prototype.parseConfig = function (config/*:AtomizerConfig*/, options/*:
                                 if (!config.hasOwnProperty('breakPoints') || !config.breakPoints.hasOwnProperty(bp)) {
                                     return;
                                 }
-                                treeo.declarations[config.breakPoints[bp]] = {};
+                                treeo.declarations[config.breakPoints[bp]] = treeo.declarations[config.breakPoints[bp]] || {};
                                 treeo.declarations[config.breakPoints[bp]][prop] = treeo.declarations[prop].replace('$' + index, value[bp]);
                             });
                             // handle default value in the custom class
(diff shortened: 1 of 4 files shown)


src-code:
cc @renatoi

renatoi:
Sorry for the delay. These changes make sense to me. The order of the media queries are important and we were not previously respecting that.

Ordering by the order the author declares the breakpoints makes sense to me since we expect authors to be declaring them in the same order of the expected CSS output.

However, my only concern is that breakpoints are declared in an object thus its [order might not be respected when we loop over this object](http://stackoverflow.com/questions/280713/elements-order-in-a-for-in-loop). The correct approach would be to declare breakpoints in an Array, but it would require a lot of refactoring, unless if we support both formats.

Since breakpoints are non-numerical, I don't think this is a big problem because I believe most browsers respect the order of definition if non-numerical. But it could happen that a browser could change its implementation in the future (order is not promised in the spec) and it could potentially break it.

Anyway, I'm leaving this decision to @src-code and @johnnorris. Other than that, I think this is a great change! Thanks!

renatoi:
@src-code @johnnorris thoughts?

src-code:
@renatoi I too have concerns about property ordering in objects, but agree that it's a problem unlikely to surface any time soon (I fear the whole web would break if an implementation were to change right now...)  At any rate, we'd be no worse off than today.  Seems fine to me to make this change. :shipit:

johnnorris:
Completely agree, it's a very valid concern, but we'll have enough warning to change it at a later date if needed, and it doesn't look like it'll happen soon. 👍🏻

On 28 Mar 2017, at 19:19, Steve Carlson <notifications@github.com<mailto:notifications@github.com>> wrote:


@renatoi<https://github.com/renatoi> I too have concerns about property ordering in objects, but agree that it's a problem unlikely to surface any time soon (I fear the whole web would break if an implementation were to change right now...) At any rate, we'd be no worse off than today. Seems fine to me to make this change. [:shipit:]

PR #291, opened 2017-05-17
Only generate CSS for media query if matches found

roderickhsiao:
Currently we are generating media query block for all breakpoints listing in options
`options.breakPoints[label]` even if we dont have any matches found (resulting in empty block)

Ideally we should only generate media query block when there are matches rule


tom76kimo:
LGTM 👍

src-code:
👍  Thanks @roderickhsiao!

Issue #296, opened 2017-06-27
Bdy (border-top+border-bottom) does not seem to work.


chhavi-khandelwal:
I used following atomic classes together.
Bdy(3px)  Bdbs(s)
But this does not seem to work. Browser does not show any rule regarding width of the border.
However, Bdw works.

thierryk:
You have a typo. It's **`BdY(3px)`**, not `Bdy(3px)`.

As a side note, the documentation contains a [whole section](https://acss.io/guides/helper-classes.html#-bd-borders-) about borders.

tinder-rhsiao:
@thierryk I think we only have `BdY` as a helper to create 1px border
https://github.com/acss-io/atomizer/blob/master/src/helpers.js#L25-L119

`BdY(3px)` probably wont work in this case

roderickhsiao:
@thierryk I think we only have `BdY` as a helper to create 1px border
https://github.com/acss-io/atomizer/blob/master/src/helpers.js#L25-L119

`BdY(3px)` probably wont work in this case

chhavi-khandelwal:
It works with BdY. Thanks!

thierryk:
cc @pankajparashar @renatoi @src-code 

Thanks @roderickhsiao. I'd copied/pasted @chhavi-khandelwal's code as-is and then simply replaced the `y` with a `Y`, leaving the extra bit :-(

As the doc shows, `BdY` does not take parameters, all helpers for borders are for a 1px border, as stated in the **Bd*** section I linked to:

> The default width of these [border] helpers is 1px as it is the most common use case. If you want to use a different width or style value, then you can either
>
>    * use standard Atomic classes, for example: `Bdw(5px) Bds(s) Bdc(#555)`

Good catch from @chhavi-khandelwal, we have typos on the [reference page](https://acss.io/reference). And it seems to be the case for most `Bd*` helpers actually (`BdX`, `BdY`, `BdT`, `BdEnd`, `BdB`, `BdStart`). I'll submit a PR.

As a side note, [Pankaj's tool](https://pankajparashar.com/atomizer-web/) shows that Atomizer plugs `Bd*` rules even if we use those classes with parameters, as in **`BdY(3px)`**. For example, `BdY(3px)` does create:

```css
.BdY {
  border-top-width: 1px;
  border-right-width: 0;
  border-bottom-width: 1px;
  border-left-width: 0;
  border-style: solid;
}
```

I don't think this could create problems though...

[...]

PR #297, opened 2017-06-28
Allowing param-to-value for Bd* classes

src-code:
Addressing issue #296 by allowing param-to-value for `Bd*` classes.

@thierryk @renatoi @roderickhsiao

### DIFF
--- a/src/rules.js
+++ b/src/rules.js
@@ -213,7 +213,7 @@ module.exports = [
         "type": "pattern",
         "name": "Border X",
         "matcher": "Bdx",
-        "allowParamToValue": false,
+        "allowParamToValue": true,
         "styles": {
             "border-__START__": "$0",
             "border-__END__": "$0"
@@ -224,7 +224,7 @@ module.exports = [
         "type": "pattern",
         "name": "Border Y",
         "matcher": "Bdy",
-        "allowParamToValue": false,
+        "allowParamToValue": true,
         "styles": {
             "border-top": "$0",
             "border-bottom": "$0"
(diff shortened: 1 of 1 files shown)


src-code:
Let me add some tests quick before we merge this...

src-code:
This is actually exposing a broken test.  Numeric custom values aren't being resolved if `allowParamToValue` is `true`.  I suppose param-to-value doesn't necessarily require the value to have a unit of measurement, since things like `line-height` don't require units.  The test name doesn't mention numeric keys being valid only when param-to-value is false, so not sure the intent here.  

I suppose we could just change Atomizer to always prefer a custom value if found, since most styles will require a unit of measurement?

```
 1) Atomizer() getCss() returns expected css value declared in custom when using numeric keys:

      AssertionError: expected '.Bdt\\(1\\) {\n  border-top: 1;\n}\n' to equal '.Bdt\\(1\\) {\n  border-top: 10px solid #ccc;\n}\n'
      + expected - actual

       .Bdt\(1\) {
      -  border-top: 1;
      +  border-top: 10px solid #ccc;
       }
      
      at Context.<anonymous> (tests/atomizer.js:701:31)
```

thierryk:
Note that things like `Bdt`are shorthand properties which would either set default styles for missing values or would require custom "variables" to work (as in this case values are space separated).

In other words, allowing passing a single parameter (not one set in the config) may have unexpected results for end users (devs).

This is ok:

.Bdt\(foo\) {
  border-top: 10px solid #ccc;
}

This is not really ok (as people may not be aware of how shorthand works or depending in which order Atomizer spits out rulesets):

```css
.Bdt\(10px\) {
  border-top: 10px;
}
```

src-code:
Ok, I'm going to close this one out since @thierryk makes a good point about the confusion shorthand could lead to.

PR #300, opened 2017-06-30
Combine pseudo class name selector if ie is off

roderickhsiao:
@src-code @thierryk 

Our default setting for Atomizer is `ie: false`, I think it is safe to combine pseudo class names selector which we previously separate them to support `< IE 8` (probably no one cares much for now).

This PR passes ie options to the `combineSelectors` function and bypass those exception rule for old IE handling.


tests/lib/jss.js
```diff
@@ -435,5 +435,41 @@ describe('JSS', function () {
             ].join('\n');
             expect(result).to.equal(expected);
         });
+        it('should combine pseudo selectors if ie options is off', function () {
```
src-code: This test seems to test both cases (on and off), so perhaps it should be broken up into two tests to be more clear what we're testing?

roderickhsiao: sure

roderickhsiao:
btw, Happy July 4 🇺🇸

src-code:
Worth noting, this has the potential to be a breaking change if people were for some reason supporting old IE but not setting the `ie` config to true.

roderickhsiao:
thats true, the assumption of of people supporting old IE will always set `ie: true` probably too strong.

roderickhsiao:
adding logic to check `undefined` and default to true for backward compatibility

roderickhsiao:
added logic to check `undefined` and default to `true` for backward compatibility, not my favorite, but let me know if you have other thoughts :)

PR #301, opened 2017-06-30
Bump lodash version

roderickhsiao:
https://github.com/lodash/lodash/releases/tag/4.0.0

@src-code @tom76kimo

### DIFF
--- a/src/lib/utils.js
+++ b/src/lib/utils.js
@@ -34,7 +34,7 @@ utils.handleMergeArrays = function (a, b) {
 // merge atomizer configs into a single config
 utils.mergeConfigs = function (configs/*:Config[]*/)/*:Config*/ {
     // TODO: Offer option to warn on conflicts
-    return _.merge.apply(null, configs.concat(utils.handleMergeArrays));
+    return _.mergeWith.apply(null, configs.concat(utils.handleMergeArrays));
 };
 
 // returns a repeated string by X amount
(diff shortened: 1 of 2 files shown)


src/lib/utils.js
```diff
@@ -34,7 +34,7 @@ utils.handleMergeArrays = function (a, b) {
 // merge atomizer configs into a single config
 utils.mergeConfigs = function (configs/*:Config[]*/)/*:Config*/ {
     // TODO: Offer option to warn on conflicts
-    return _.merge.apply(null, configs.concat(utils.handleMergeArrays));
+    return _.mergeWith.apply(null, configs.concat(utils.handleMergeArrays));
```
src-code: @roderickhsiao Any particular reason to make this change, since we're not using the customer argument that differentiates `mergeWith` from `merge`?

roderickhsiao: ha, the unit tests fails if stay with `merge`

src-code: That's worrisome!

roderickhsiao: I just found out that we did pass the customizer `utils.handleMergeArrays` in the arguments 

`_.merge.apply(null, configs.concat(utils.handleMergeArrays))`

thats why we need to change to `mergeWith`

src-code: Ah, I see, and in 3.x `merge` took that arg as well.  Looks good then.

PR #303, opened 2017-10-05
Add support for general sibling combinator

renatoi:
Based on a gitter conversation, we currently do not support the general sibling selector which can be very useful in many situations. @thierryk @src-code Please review and see if it makes sense. Also, we need to update the docs once this is accepted.


src-code:
@renatoi Whoa!  You're back!  Will check this out...

src-code:
We never updated the docs, @renatoi 😄

PR #304, opened 2017-10-09
Generate the flex property with custom values from Flx(...)

snyamathi:
@src-code @redonkulus @sfj2 Currently, you cannot generate the required atomic css for the the `flex` property with the exception of `auto` and `none`.

This PR adds `initial` and also the ability to use a value ( 1, 2, 3, 10em, 30px, etc ).

`flex` is a shorthand property, so 

- `Flx(2)` would generate `flex: 4` which is shorthand for `flex-grow: 4`
- `Flx(10px)` would generate `flex: 10px` which is shorthand for `flex-basis: 10px`

We support other shorthand properties, such as `M(0)` and `P(0)` already, so this isn't entirely new ground.

I'm also fixing a few lint errors that I found.

### DIFF
--- a/src/lib/jss.js
+++ b/src/lib/jss.js
@@ -46,7 +46,6 @@ JSS.flattenSelectors = function (newJss/*:Jss*/, jss/*:Jss*/, parent/*:string*/)
 JSS.extractProperties = function (extracted/*:Extracted*/, jss/*:JssFlat*/, block/*:string*/)/*:Extracted*/ {
     var props;
     var prop;
-    var extract;
 
     block = block || 'main';
 
--- a/src/rules.js
+++ b/src/rules.js
@@ -1122,6 +1122,7 @@ module.exports = [
         },
         "arguments": [{
             "a": "auto",
+            "ini": "initial",
             "n": "none"
         }]
     },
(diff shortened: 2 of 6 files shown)


src-code:
I don't think this is a good idea, since `Flx` (`flex`) is a shorthand class, meant to take multiple values separated by spaces, which can't be done in Atomic without a custom value.  If you're only passing a single value to `Flx`, you should just use the exact flex style you're trying to affect (eg instead of `Flx(2)`, use `Flxg(2)`)

src-code:
@thierryk @renatoi fyi

src-code:
BTW, `M(0)` and `P(0)` is a little different, since it's a single value applied to all sides of a box, so param-to-value makes more sense.  

`Flx` is more analogous to `Bd`, which is shorthand for `border-color`, `border-width`, and `border-style`, and therefore doesn't accept param-to-value in Atomizer.

snyamathi:
The suggestion to use the specific property makes sense.  Multiple values do get very confusing, even in the case of border / margin, especially with 3 values, so I can understand the hesitation here.  I think we can get by with using Flxg then ( @sfj2 )

sfj2:
@src-code @snyamathi so `flex:1` and `flex-grow: 1`,  are not exactly the same, when using `flex:1` chrome would also add `flex-shrink:1`, so when using flxg(1) we would also have to do flxs(1) too. so it's just convenient to do flex: 1, example -

animation

Issue #316, opened 2017-12-21
:link, :visited, :hover, and :active pseudo-classes should be ordered (in that sequence).


chhavi-khandelwal:
```HTML

<button class="Bgc(#e6b105):h   Bgc(#000000):a  Bgc(#eeeff1)">12345678</button>

```

```Generated CSS
.Bgc\(\#000000\)\:a:active {
  background-color: #000000;
}
.Bgc\(\#e6b105\)\:h:hover {
  background-color: #e6b105;
}
.Bgc\(\#eeeff1\) {
  background-color: #eeeff1;
}
```
While clicking the button, active state triggers and hover state also applies. But in generated CSS active state CSS is overriden by hover state.
Correct order should be: 
```
link
visited
hover
active
```
Reference doc: https://css-tricks.com/almanac/selectors/a/active/

thierryk:
@src-code @renatoi @redonkulus This is a good catch from @chhavi-khandelwal

tomassentkeresty:
Hello, you can also consider putting :focus before :hover and :active.

src-code:
Thanks everyone!  This will land in 3.5.0

PR #320, opened 2018-02-24
Fix pseudo-classes order

idmytro:
Fix this https://github.com/acss-io/atomizer/issues/316 bug


redonkulus:
@idmytro just seeing this now. Can you add a test for the new code?

PR #330, opened 2018-11-02
Report invalid match value.

orgoj:
Report invalid matched value for better find error in html.

### DIFF
--- a/src/atomizer.js
+++ b/src/atomizer.js
@@ -215,6 +215,10 @@ Atomizer.prototype.parseConfig = function (config/*:AtomizerConfig*/, options/*:
             // parse them and return a valid value
             values = values.split(',').map(function (value, index) {
                 var matchVal = Grammar.matchValue(value);
+                if (matchVal == null) {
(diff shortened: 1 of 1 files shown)


src-code:
@orgoj Thanks for this, could you please add a test to go along with your code change?

src/atomizer.js
```diff
@@ -215,6 +215,10 @@ Atomizer.prototype.parseConfig = function (config/*:AtomizerConfig*/, options/*:
             // parse them and return a valid value
             values = values.split(',').map(function (value, index) {
                 var matchVal = Grammar.matchValue(value);
+                if (matchVal == null) {
```
src-code: Should use `===`

src/atomizer.js
```diff
@@ -215,6 +215,10 @@ Atomizer.prototype.parseConfig = function (config/*:AtomizerConfig*/, options/*:
...
+                    console.warn('Error: Invalid value "' + value + '".');
```
src-code: It'd be helpful if the error message also included the selector so the user can more easily identify the problem

Issue #328, opened 2018-09-28
Support minimatch matching file


roderickhsiao:
Will be good to support CLI minimatch so that when using recursive options we can still decide which file we want to scan.

PR #333, opened 2018-11-27
Support excluding file from CLI

roderickhsiao:
Support minimatch pattern to exclude file being scanned.

https://github.com/acss-io/atomizer/issues/328

This is useful when you are using recursive to scan the whole folder but you want to exclude file such as test file. 

I'm not sure should we support `include`, feel like it will become a bit complicated 

@src-code @redonkulus


bin/atomizer
```diff
@@ -10,12 +10,14 @@
...
 process.title = 'atomizer';
 
-var path = require('path');
-var fs = require('fs');
 var chalk = require('chalk');
+var fs = require('fs');
+var minimatch = require('minimatch');
+var path = require('path');
 var program = require('commander');
-var Atomizer = require('../src/atomizer');
```
src-code: @roderickhsiao Don't think you meant to remove this

bin/atomizer
```diff
@@ -70,9 +74,19 @@ function parseFile (file, recursive, dir) {
...
+              isExcluded = some(program.exclude, function excludeFile(value) {
+                return minimatch(filepath, value, { matchBase: true });
+              })
+            }
+            if (!isExcluded) {
+              if (!program.quiet) console.warn('Parsing file ' + chalk.cyan(relative) + ' for Atomic CSS classes');
+              fileContents = fs.readFileSync(filepath, {encoding: 'utf-8'});
+              classNames = atomizer.findClassNames(fileContents);
+            } else {
+              if (!program.quiet) console.warn('Excluding file ' + chalk.cyan(relative) + ' for Atomic CSS classes');
```
src-code: Awesome, you picked up the `quiet` change I just merged 💯

roderickhsiao: I think a better way is to just wrap `console.warn` and have one check, but I'm lazy ...

src-code: Probably true, but this is fine :)

bin/atomizer
```diff
@@ -10,12 +10,15 @@
...
 
-var path = require('path');
-var fs = require('fs');
 var chalk = require('chalk');
+var fs = require('fs');
+var minimatch = require('minimatch');
+var path = require('path');
 var program = require('commander');
+var some = require('lodash/some');
+var union = require('lodash/union');
```
src-code: @roderickhsiao This doesn't appear to be used?

roderickhsiao: oh I think it got removed from previous PR

roderickhsiao: oh in `parseFiles`

bin/atomizer
```diff
@@ -47,11 +51,15 @@ if (process.argv.length <= 2) {
 
 var atomizer = new Atomizer({ verbose: !!program.verbose });
 
+function warn(output) {
+  if (!program.quiet) console.warn(outout);
```
src-code: typo

src-code: `console.warn` can take N arguments, which is used in one place.  Should just take `arguments` and pass via `apply()`

src-code: Correction - we're not actually using extra args currently.  I read line 96 wrong!

roderickhsiao: :) I think it no harm just take `arguments` so if later we want to use it will be easier

src-code:
@roderickhsiao Hmm, something doesn't seem to work as expected with certain args.  If I have two files, `test.txt` and `test.excl`:

**test.txt**
```
<div class="Bd(0)"></div>
```

**test.excl**
```
<div class="D(n)"></div>
```

And run the following command:
```
 $ atomizer test.* -e *.excl 
Excluding file test.excl for Atomic CSS classes
Parsing file test.txt for Atomic CSS classes

.Bd\(0\) {
  border: 0;
}
```
It works right.  But if I do:

```
$ atomizer test.* -e *.txt
Parsing file test.excl for Atomic CSS classes
Parsing file test.txt for Atomic CSS classes
Parsing file test.txt for Atomic CSS classes

.Bd\(0\) {
  border: 0;
}
.D\(n\) {
  display: none;
}
```
It's not working right...

src-code:
Ah, if I quote the last expression it works right!

roderickhsiao:
you mean `$ atomizer test.* -e "*.txt"` tghis works?

src-code:
Correct

roderickhsiao:
```
bin/atomizer test.* -e *.txt
Parsing file test.excl

 for Atomic CSS classes
Excluding file test.txt for Atomic CSS classes

.D\(n\) {
  display: none;
}
```

When I log locally, I see the filepath has lots of empty string for some reason, after `trim()` it I can see it working.

Its's wired that when I tried `atomizer examples/html/*.html -e *-0.html` I didn't see the space issue.

========== file path start ==========
/atomizer/test.excl


========== file path end ==========
========== file path start ==========
/atomizer/test.txt
========== file path end ==========

roderickhsiao:
ignore my previous comment, I created the file with space.
Seems work for me for both case.

My result 

```
bin/atomizer test.* -e *.excl 
Excluding file test.excl for Atomic CSS classes
Parsing file test.txt for Atomic CSS classes

.Bd\(0\) {
  border: 0;
}
```


```
bin/atomizer test.* -e *.txt
Parsing file test.excl for Atomic CSS classes
Excluding file test.txt for Atomic CSS classes

.D\(n\) {
  display: none;
}

```

src-code:
I'm running against files in the root atomizer directory, but if I move the files to a subdir and run the commands, then it works correctly.  So there's something going on with how it's resolving paths I guess

src-code:
@roderickhsiao Ah, I see what's going on.  It's because `-e *.txt` gets expanded by the shell, which screws up the args atomizer receives (it treats all the expanded files as "files to parse" rather than exclude).  If you do `--exclude=*.txt` it doesn't get expanded, so it works fine.  Same for quoting the string passed to `-e`.  I think the best solution is to eliminate the `-e`, and only offer `--exclude=`

roderickhsiao:
ah! Thanks for debugging! okay, let me just remove the `-e`

README.md
```diff
@@ -40,13 +40,15 @@ Usage: atomizer [options] [path]
...
     --ie                                adds old IE hacks to the output.
+    --exclude [pattern]                 pattern to exclude file to be scanned
     --verbose                           show additional log info (warnings).
 ```
 
 Example:
 
 ```
 atomizer -o atomic.css ./site/*.html
+atomizer -o atomic.css ./site/*.html --exclude *-0.html
```
src-code: Should be `--exclude=*-0.html`

roderickhsiao: when you need morning coffee

README.md
```diff
@@ -40,13 +40,15 @@ Usage: atomizer [options] [path]
     -H, --helpersNamespace [namespace]  adds the given namespace to all helper selectors.
     --rtl                               swaps `start` and `end` keyword replacements with `right` and `left`.
     --ie                                adds old IE hacks to the output.
+    --exclude [pattern]                 pattern to exclude file to be scanned
```
src-code: Probably should add the `=` here too

README.md
src-code: Same here!

README.md
src-code: Actually all of the README examples need it, I'll do this seperately

roderickhsiao: @src-code it seems like adding `=` then it doesn't work anymore.

program.exclude is `undefined`

https://www.npmjs.com/package/commander

if I remove equal then it works
`.option('--exclude [pattern]', 'excluded file pattern', collect, [])`

src-code: Ah, interesting.  Ok, will merge your other PR

PR #334, opened 2018-11-28
Sort pseudo classes

roderickhsiao:
Addressed https://github.com/acss-io/atomizer/issues/316

credit to https://github.com/acss-io/atomizer/pull/320
@idmytro

PR #339, opened 2018-12-05
Keep default original sort logic if match name are different 

roderickhsiao:
This is a regression caused by #334 

We are sorting all class name again after the initial sort, the second sort for pseudo class should only care about the class name with the same atomic selector

let's say

`C(#fff):a` and `C(#000):h` should be sorted by the pseudo order but not
`Op(#fff):a and `C(#000):h` (it should remain the same alphabetical order as the initial sort)

This cause issue that some CSS weight being changed cause the order change.

### DIFF
--- a/src/atomizer.js
+++ b/src/atomizer.js
@@ -144,7 +144,17 @@ Atomizer.prototype.sortCSS = function (classNames /*string[]*/) {
                 return _.includes(value, pseudoClass);
             });
           }
-        return getMatchedIndex(a) - getMatchedIndex(b);
+        const aMatches = Grammar.matchValue(a);
+        const bMatches = Grammar.matchValue(b)
+        const aIndex = getMatchedIndex(a);
+        const bIndex = getMatchedIndex(b);
+
+        // remain same default sort logic
+        if (aMatches.named !== bMatches.named) {
+          return a.localeCompare(b);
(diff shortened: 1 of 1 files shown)


tests/atomizer.js
```diff
@@ -976,38 +976,87 @@ describe('Atomizer()', function () {
...
-              'P(20px)',
-              'T(0)',
-              'C(#fff):li',
-              'C(#123):vi',
-              'Scale(200%):f',
-              'Op(1):h',
-              'C(#000):a'
+                'C(#fff):li',
+                'C(#123):vi',
+                'C(#000):a',
```
roderickhsiao: 1. same class with different pseudo class should be sorted

tests/atomizer.js
```diff
@@ -976,38 +976,87 @@ describe('Atomizer()', function () {
...
+                'D(b)',
+                'Mb(10px)',
+                'Op(1):h',
+                'P(20px)',
+                'Scale(200%):f',
+                'T(0)'
```
roderickhsiao: other matches should be alphabetical order

roderickhsiao:
are we able to release a version for the fix :)?

roderickhsiao:
Haha you guys can test more :)

PR #341, opened 2018-12-24
feat(alpha-in-custom): added support for alpha in custom value

### LINKED ISSUE #340: Question - Alpha transparency
```
// acss.config.js
module.exports = {
  'custom' : {
     black: '#000',
     black015: '#000.15'
  }
}
```

```
// .html
class="Bdbc(#000.15) Bdbc(black015) Bdbc(black)"
``` 

```
// generated css file
.Bdbc\(\#000\.15\) {
  border-bottom-color: rgba(0,0,0,.15);
}
.Bdbc\(black015\) {
  border-bottom-color: #000.15;
}
.Bdbc\(black\) {
  border-bottom-color: black;
}
```

Hi @renatoi, @src-code, @thierryk, 

Happy Holidays!

With respect to the above example I am trying to avoid any random values in the html and rely on the `custom` hash of the config file. Unfortunately I am unable to do it correctly.
I am wondering if there is a better way tp generate rules with alpha transparency from predefined custom properties.

Any pointer would be helpful.

Thank you!

sarbbottam:
fixes #340

```js
// acss.config.js
module.exports = {
  'custom' : {
     black: '#000'
  }
}
```
```html
// .html
class="Bg(black.15)"
```
```css
// generated css file
/* current behavior */
.Bg\(black\.15\) {
  background: #000;
}
/* future behavior if this RB is merged */
.Bg\(black\.15\) {
  background: rgba(0,0,0,.15);
}
```

Hello team, it would be great to get some thoughts and feedback for this PR. Thank you for your time.

Best!

### DIFF
--- a/src/atomizer.js
+++ b/src/atomizer.js
@@ -170,6 +170,7 @@ Atomizer.prototype.parseConfig = function (config/*:AtomizerConfig*/, options/*:
     var warnings = [];
     var isVerbose = !!this.verbose;
     var classNames = config.classNames;
+    var alphaInName = '';
 
     if (!_.isArray(config.classNames)) { return tree; }
     options = options || {};
@@ -315,6 +316,24 @@ Atomizer.prototype.parseConfig = function (config/*:AtomizerConfig*/, options/*:
                         // as value
                         else if (config.custom.hasOwnProperty(matchVal.named)) {
                             value = config.custom[matchVal.named];
+
+                            if (Grammar.matchValue(value).hex === value && matchVal.named !== matchVal.input) {
+                                alphaInName = matchVal.input.replace(matchVal.named, '');
+                                if (alphaInName && alphaInName.match(/\.\d{1,2}/)) {
+                                    rgb = utils.hexToRgb(value);
+                                    value = [
+                                        'rgba(',
+                                        rgb.r,
+                                        ',',
+                                        rgb.g,
+                                        ',',
+                                        rgb.b,
+                                        ',',
+                                        alphaInName,
+                                        ')'
+                                    ].join('');
+                                }
+                            }
                         }
                         // we have custom but we could not find the named class name there
                         else {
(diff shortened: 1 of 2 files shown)


sarbbottam:
Happy New Year!

It would be great to have few 👀for this PR.

cc/ @renatoi @src-code @thierryk @tom76kimo @pankajparashar-zz 

Thanks!

sarbbottam:
@src-code could you share your thoughts? Thanks!

src-code:
Since custom values aren't currently interpreted in general, I think we should tread carefully here.  I'd suggest going with a custom value like `black015: 'rgb(0,0,0,.15)'` instead

sarbbottam:
>`black015: 'rgb(0,0,0,.15)'`

@src-code Thanks for the review that was my initial utilization.

We currently use SASS  accross the organization and one of the use-case that it is facilitation is to control the usage of colors accross the organization.

The colors are currently controlled by using the `get-color()` function provide the base SASS utilities. The `get-color()` function accepts two arguments, the mandatory first argument is the name of the whitelisted color and and optional second argument the opacility like so:

```scss
...
background-color: get-color(black, .15) // background-color: rbga(0, 0, 0,.15) 
...

```

The motivation behind this PR was to create a `acss.config.js` like so:

```
// acss.config.js
const colors = require('./colors');

const config = {
  custom: {
    ...
  },
};
config.custom = Object.assign(config.custom, colors);

module.exports = config;
```

The `colors.js` would be exporting an hash of approved `color name` and their corresponding `hex` pair.

```js
const approvedColors = require('org/colors.js');
const colors = _createColorsHash(approvedColors);
module.exports = colors;
```

So that one could simply use `color.alpha` combination with out an explicit entry in the `acss.config.js`.

---

Otherwise, there needs to be an explit entry for any new `color.alpha` combination.

```
// acss.config.js
black015: 'rgb(0,0,0,.15)',
black050: 'rgb(0,0,0,.50)'
...
```

However it would be better to have 

```
// acss.config.js
black: '#000'
...
```
And any transparency combination value can be resulted just with the entry of the color.

Let me know if you think this could be valid requirement for `atomizer` and we can work towards a better solution.

PR #342, opened 2018-12-24
feat(bin/atomizer): writing to outfile only if the content has changed

sarbbottam:
In my existing corporate dev workflow there is already a watcher watching the entire directory for any changes and triggering a build. I do not have much control over the mentioned watcher and thought about this change to write to the output file only if the content has changed since the previous write.

Looking forward to your feedback, suggestions. Thanks!

### DIFF
--- a/bin/atomizer
+++ b/bin/atomizer
@@ -160,12 +160,18 @@ content = atomizer.getCss(config, options);
 // Output the CSS
 var outfile = program.outfile;
 if (outfile) {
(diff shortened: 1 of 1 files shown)


sarbbottam:
Happy New Year!

It would be great to have few 👀for this PR.

cc/ @renatoi @src-code @thierryk @tom76kimo @pankajparashar-zz 

Thanks!

Issue #353, opened 2019-09-03
Jc(se): space-evenly not supported


fidian:
`Jc(sb)` works as expected, as does `Jc(sa)`, but `Jc(se)` does not exist. It's [widely supported](https://caniuse.com/#feat=justify-content-space-evenly). The CSS 3 spec says it's a member of [`content-distribution`](https://www.w3.org/TR/css-align-3/#typedef-content-distribution), which [applies to `justify-content`](https://www.w3.org/TR/css-align-3/#propdef-justify-content) according to W3C Working Draft, 6 December 2018. Can this get added to atomizer so I don't need to add a custom value as was done in #322?

src-code:
Added in v3.6.1

fidian:
Thanks! It works like a charm.

PR #354, opened 2019-09-11
Support more justify content attribute

roderickhsiao:
https://developer.mozilla.org/en-US/docs/Web/CSS/justify-content

solve #353


tom76kimo:
Don't have permission 😢

redonkulus:
Need @src-code to publish it

src-code:
Published in v3.6.1

roderickhsiao:
Thanks!!

Issue #355, opened 2019-09-27
Add support for font-variant-numeric


carystanley:
I don't see support for `font-variant-numeric`

https://developer.mozilla.org/en-US/docs/Web/CSS/font-variant-numeric

PR #356, opened 2019-09-30
Support more font-variant attributes

ArvinH:
## Goal

Support syntax like `'Fv(sc,slz)'` could be transpiled to `font-variant: small-caps slashed-zero;',` with ```config.custom: {
                    'sc': 'small-caps',
                    'slz': 'slashed-zero'
                },```
## Relevant issue 

https://github.com/acss-io/atomizer/issues/355

## MDN Docs

https://developer.mozilla.org/en-US/docs/Web/CSS/font-variant

### DIFF
--- a/tests/atomizer.js
+++ b/tests/atomizer.js
@@ -899,7 +899,11 @@ describe('Atomizer()', function () {
         it ('properly handles classnames with optional arguments', function () {
             var atomizer = new Atomizer();
             var config = {
-                classNames: ['Skew(90deg)', 'Skew(90deg,45deg)', 'Bdsp(1em)', 'Bdsp(2em,33%)']
+                custom: {
+                    'sc': 'small-caps',
+                    'slz': 'slashed-zero'
+                },
+                classNames: ['Skew(90deg)', 'Skew(90deg,45deg)', 'Bdsp(1em)', 'Bdsp(2em,33%)', 'Fv(sc,slz)']
             };
             var expected = [
                 '.Bdsp\\(1em\\) {',
@@ -907,6 +911,9 @@ describe('Atomizer()', function () {
                 '}',
                 '.Bdsp\\(2em\\,33\\%\\) {',
                 '  border-spacing: 2em 33%;',
+                '}',
+                '.Fv\\(sc\\,slz\\) {',
+                '  font-variant: small-caps slashed-zero;',
                 '}',
                 '.Skew\\(90deg\\) {',
                 '  transform: skew(90deg);',
--- a/src/rules.js
+++ b/src/rules.js
@@ -1526,7 +1526,7 @@ module.exports = [
         "matcher": "Fv",
         "allowParamToValue": false,
         "styles": {
-            "font-variant": "$0"
+            "font-variant": "$0 $1"
         },
         "arguments": [{
             "n": "normal",


tom76kimo:
Should we add the possible value like `no-common-ligatures`, `common-ligatures`, etc. to `arguments`?

arvinhv:
🤔 After CSS3, there are four properties could be used by this shorthand syntax and each one has various values. Instead of adding more arguments, I was thinking about whether we should create new rules for them accordingly. But anyway, I extend its capability of setting up two attributes at a time for now.

Infact, for #355, I think we've already supported it by using custom value, browsers who are supported these shorthand syntax would understand.

ArvinH:
🤔 After CSS3, there are four properties could be used by this shorthand syntax and each one has various values. Instead of adding more arguments, I was thinking about whether we should create new rules for them accordingly. But anyway, I extend its capability of setting up two attributes at a time for now.

In fact, for #355, I think we've already supported it by using custom value, browsers who are supported these shorthand syntax would understand.

tom76kimo:
@redonkulus @src-code

roderickhsiao:
LGTM, I think its fine to keep the variable as no one really filing issue about other missing arguments. We can add it later.

tom76kimo:
@roderickhsiao Cool, @src-code how do you think? If it's ok, could you help review and merge?

ArvinH:
@src-code thanks for pointing out, I guess I misunderstood the meaning of `allowParamToValue`, will do a test for it.

ArvinH:
@src-code 
Maybe I didn't convey my purpose clearly enough, so I add an example to PR description and add a simple test for it. (please correct me if I need more test or should put it to somewhere else.)

Also, seems like `allowParamToValue` only affect the result when `matchVal` is a number
https://github.com/acss-io/atomizer/blob/master/src/atomizer.js#L252-L253
and for `font-variant`, the value won't be a number, it'll be `named` for its `matchVal`.

tests/atomizer.js
```diff
@@ -899,7 +899,11 @@ describe('Atomizer()', function () {
         it ('properly handles classnames with optional arguments', function () {
             var atomizer = new Atomizer();
             var config = {
-                classNames: ['Skew(90deg)', 'Skew(90deg,45deg)', 'Bdsp(1em)', 'Bdsp(2em,33%)']
+                custom: {
+                    'sc': 'small-caps',
```
src-code: `sc` is already defined as a predefined argument for `font-variant`, so you wouldn't want to confuse things by having a custom value with the same name.

ArvinH:
You made a good point, I wasn't thinking clearly, syntax like `Fv(sc, slz)` is not helping (probably only can let user reuse custom value, like `Fv(sc,slz)` and `Fv(ord,slz)` where `ord = ordinal`)

In terms of reusability and the principle stated [here](https://acss.io/guides/shorthand.html), I agree with you, shouldn't go in this way.

I can add rules for each `font-variant-*` attribute.

src-code:
> On second thought, attribute like font-variant-alternates and font-variant-numeric, they both have many acceptable values and even can have two values at a time, if we create arguments for each possible value, that'd be unrealistic for users to know/remember them and have to look up those matchers every time, at the end they might just create custom value instead...perhaps we should keep this as it.

I'm fine either way, I think there's value in offering classes rather than forcing people into custom values as they tend to be more easily reused especially in larger teams, but I'll leave that up to whoever wants to open a PR. :)

PR #357, opened 2019-10-02
feat: support object-fit

snyamathi:
@src-code Open to renaming the shorthand from `Objf` to anything else but this PR is to support the [object-fit](https://developer.mozilla.org/en-US/docs/Web/CSS/object-fit) property.

I've seen the values such as `contain`, `cover` used elsewhere but I didn't see an existing example for `scale-down` so I shorthanded it as `sd`.  Open to changing that as well.


src/rules.js
```diff
@@ -1869,6 +1869,27 @@ module.exports = [
     },
     /**
     ==================================================================
+    OBJECTS
```
snyamathi: I gave this a more generic name, objects, since there are other properties like [object-position](https://developer.mozilla.org/en-US/docs/Web/CSS/object-position) which may be covered later on

snyamathi:
The build for node 4 has failed - though node 4 was [EOL](https://github.com/nodejs/Release#end-of-life-releases) a year and a half ago

src/rules.js
src-code: I think this is fine, we try to follow Emmet syntax but Emmet doesn't seem to support these yet

Issue #358, opened 2019-10-25
UnhandledPromiseRejectionWarning: TypeError: Cannot read property 'number' of null


li-yuening:
at /Users/liyuening/.config/yarn/global/node_modules/atomizer/src/atomizer.js:252:30
    at Array.map (<anonymous>)
    at Atomizer.<anonymous> (/Users/liyuening/.config/yarn/global/node_modules/atomizer/src/atomizer.js:248:40)
    at Array.forEach (<anonymous>)
    at Atomizer.parseConfig (/Users/liyuening/.config/yarn/global/node_modules/atomizer/src/atomizer.js:181:16)
    at Atomizer.getCss (/Users/liyuening/.config/yarn/global/node_modules/atomizer/src/atomizer.js:488:17)
    at buildAtomicCss (/Users/liyuening/.config/yarn/global/node_modules/atomizer/bin/atomizer:142:24)
    at triggerBuild (/Users/liyuening/.config/yarn/global/node_modules/atomizer/bin/atomizer:178:5)
    at FSWatcher.<anonymous> (/Users/liyuening/.config/yarn/global/node_modules/atomizer/bin/atomizer:221:13)
    at FSWatcher.emit (events.js:198:13)
(node:10424) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 1)
(node:10424) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.

Version: 3.6.2
Node Version: v10.16.0
Operations: atomizer --config="./atomizer-config.js" --outfile="./www/static/sales/css/v3-atomic.css" --watch "./**/v3_main_template.html"
Editor: VSCode with autosave

Error just came up. I cannot figure out why. It happened twice so far within 2 hours using of atomizer. I have to ctrl + c and run the command again.

I'd like to provide more information if there's any.

li-yuening:
I found that once I typed a hyphen, atomizer would crash and output something as above.

eg: End(-)

tom76kimo:
Interesting, when will you use a single hyphen as a value? But yes, we can make the code more defensive here https://github.com/acss-io/atomizer/blob/master/src/atomizer.js#L249 . The `matchVal` could be a `null` in this case.

li-yuening:
[...]

PR #359, opened 2019-12-03
Handle cases would break the code

tom76kimo:
Based on https://github.com/acss-io/atomizer/issues/358, though I think it's a rare case, we can still add defensive code here to prevent possible bug.
How do you think? @src-code @redonkulus


tom76kimo:
@redonkulus @src-code

PR #363, opened 2020-01-09
Allow custom-params to include commas. Resolves #362.

### LINKED ISSUE #362: custom-param should allow commas
Consider, for example, the `Tsh` function:

```
Tsh(<custom-param>)
text-shadow: value
```

I would like to parameterize my custom values by size and color, e. g. `Tsh(sm,gray300)`, by adding something like the following to my config:

```
{
  custom: {
    "Tsh(sm,gray300)": "1px 1px #ccc"
  }
}
```

This is currently not possible because the `parseConfig` function does not check for `config.custom.hasOwnProperty("Tsh(sm,gray300)")` before splitting `sm,gray300` and treating `sm` and `gray300` as separate values.

nsaunders:
Apologies for the extra diffs, but it was necessary to indent some existing code.

### DIFF
--- a/src/atomizer.js
+++ b/src/atomizer.js
@@ -243,89 +243,89 @@ Atomizer.prototype.parseConfig = function (config/*:AtomizerConfig*/, options/*:
         if (match.atomicValues) {
             values = match.atomicValues;
 
-            // values can be separated by a comma
-            // parse them and return a valid value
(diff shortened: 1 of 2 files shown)


nsaunders:
Closing due to lack of interest.

PR #369, opened 2020-03-31
Bail out when null values are encountered

src-code:
When invalid classes are encountered such as `P(a,0,c)` which result in null values being generated for some args (they are not custom values), we need to short-circuit the processing of the rest of the values.  Can't short-circuit a `forEach` loop, but can check for `treeo.declarations` being null before trying to reference it further.

Background - This issue came up because of a recent change to Marko templating system.  Marko generates code that looks a little like Atomic classes, but is actually obfuscated function calls.  Normally these would be ignored due to the lack of valid "values" in the string, but they recently switched away from serializing `null` to using a `0` value, which looks "valid" to Atomizer. 

https://github.com/marko-js/marko/pull/1535


redonkulus:
Thanks!

PR #370, opened 2020-04-03
chore: add typescript types

snyamathi:
This PR adds typescript types for anyone using `atomizer` in a typescript project.

We could certainly lock down some of the methods to be private, but I don't see any harm.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -10,14 +10,7 @@ declare module 'atomizer' {
     }
 
     export interface AtomizerRule {
-        allowParamToValue: boolean;
-        arguments?: Record<string, string>[];
-        id?: string;
-        matcher: string;
-        name: string;
-        shorthand?: boolean;
-        styles: Record<string, string>;
-        type: string;
+        [key: string]: any;
     }
 
     export interface CSSOptions {
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,6 +1,6 @@
 declare module 'atomizer' {
     export interface AtomizerConfig {
-        custom?: Record<string, string>;
+        custom?: Record<string, string | Record<string, string>>;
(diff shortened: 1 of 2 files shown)


index.d.ts
```diff
@@ -0,0 +1,49 @@
+declare module 'atomizer' {
+    export interface AtomizerConfig {
+        custom?: Record<string, string>;
```
src-code, line 3: Custom values can be objects too: https://github.com/acss-io/atomizer/blob/master/tests/atomizer.js#L319-L326

snyamathi: updated https://github.com/acss-io/atomizer/pull/370/commits/67373c462bf230f058ee1e33f50e003d98f417f2

index.d.ts
```diff
@@ -0,0 +1,49 @@
...
+        breakPoints?: Record<string, string>;
+        classNames?: string[];
+    }
+
+    export interface AtomizerOptions {
+        verbose?: boolean;
+    }
+
+    export interface AtomizerRule {
```
src-code, line 12: Need to distinguish between rules and helpers, helpers have different values, for instance: https://github.com/acss-io/atomizer/blob/master/src/helpers.js#L186-L211

snyamathi: I'll just relax the rules as `any` for now and we can enhance later on - it seems like it's a pretty complex object that could be broken down into smaller parts if we want to flesh this out more.

src-code: Rules and helpers are processed through the same APIs, just have different shape

snyamathi: updated https://github.com/acss-io/atomizer/pull/370/commits/a315f653e99b60c80408443f7a0a65b01dc6e987

PR #374, opened 2020-07-10
fix: false positives

snyamathi:
@src-code Reduce false positives by _only_ allowing `=` with `class=` otherwise we would match minified js such as `foo=D(b)`

### DIFF
--- a/src/lib/grammar.js
+++ b/src/lib/grammar.js
@@ -63,7 +63,7 @@ function flatten(obj) {
 
 // regular grammar to match valid atomic classes
 var GRAMMAR = {
-    'BOUNDARY'      : '(?:^|\\s|=|"|\'|\`|\{|\})',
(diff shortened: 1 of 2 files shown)


src-code (approved):

renatoi:
Some frameworks such as React uses `className` instead of class (reserved keyword in JS). I'd modify this to also accept `className`.

snyamathi:
`    'BOUNDARY'      : '(?:^|\\s|=|"|\'|\`|\{|\})',`


@renatoi in that case, the quote proceeding the class name eg `className='foo'` would match. This is really more of a html/marko specific thing from https://github.com/acss-io/atomizer/pull/373 where we have a document such as

```html
<div class=D(n)>Hello World</div>
```

Which is valid HTML **without the quotes** but for something like `className` the quotes would be required since that's not valid.

renatoi:
ah makes sense! 👍

PR #379, opened 2020-10-17
Upgrade dependencies

itssumitrai:
Upgrade `mocha`
Use `nyc` instead of `istanbul`
remove `object-assign` in favor of `Object.assign`
Upgrade minimum node version to v10 since `mocha` breaks in `<10`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,9 +5,6 @@ matrix:
 node_js:
   - "11"
   - "10"
-  - "8"
-  - "6"
-  - "4"
 after_success:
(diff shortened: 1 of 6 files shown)


.travis.yml
```diff
@@ -5,9 +5,6 @@ matrix:
 node_js:
   - "11"
   - "10"
```
redonkulus, line 7:
```suggestion
node_js:
  - "12"
  - "10"
```

redonkulus (approved):
thanks!

PR #382, opened 2021-02-06
Added a "Third Party Extensions" section

thierryk:
I added VSCode eand Chrome extensions below the "Plugins" section as I could not find a place where those resources were consolidated.
I'm just wondering if we should "explicitly tag" these resources as third party resources (?).


redonkulus:
If the core team does not maintain them then the header should be Third Party Extensions.

redonkulus (approved):

Issue #384, opened 2021-02-12
RFC: Optional specificity bump for media queries


src-code:
One issue that sometimes comes up when using Atomic CSS produced by Atomizer is that media queries carry the same weight as their non-MQ'ed classes.  This can cause issues when multiple Atomic CSS stylesheets need to be combined on the page dynamically where the whole of the CSS can not be known in advance, for example through lazy-loading of modules from multiple servers.  In this scenario, source-order determines which style will "win":

 For example:

*first.css*
```css
.C\(white\) {
  color: white;
}
@media(min-width:1px) {
  .C\(red\)--sm {
    color: red;
  }
}
```

*second.css*
```css
.C\(white\) {
  color: white;
}
```

*index.html*
```html
<link rel="stylesheet" href="first.css">
<link rel="stylesheet" href="second.css">
<div class="C(white) C(red)--sm">This text will be blue even though the media query matches</div>
```

I'd like to propose adding an option to Atomizer that makes it possible to bump the specificity of media queries to make this scenario less likely to happen.  This could be done in a couple of ways - we could either have a single way to bump specificity, or we could offer optional levels of specificity.  

For example, we could bump specificity a small amount by using a technique such as the "double-selector" (credit: @thierryk):

```css
/* specificity is 0,0,1,0 */
.C\(white\) {
  color: white;
}

@media(min-width:1px) {
  /* specificity is 0,0,2,0 */
  .C\(red\)--sm.C\(red\)--sm {
    color: red;
  }
}
```

This technique would similarly be applied even with things like descendent selector classes, with or without namespaces (which apply `!important` to bump the specificity of the descendent selector when namespaces are in use.)

If double-selector isn't deemed sufficient, we could offer an additional option to bump specificity with a hammer like `!important`. 🔨  I *think* bumping specificity the minimum amount is probably best and sufficient, but unsure if there would be valid cases where you'd want to increase specificity further.  

I'd propose the option to Atomizer could be something like the following:

```bash
atomizer -c config.json index.html -n  \#atomic --bump-mq
```

Or, two options:

```bash
atomizer -c config.json index.html -n  \#atomic --bump-mq
atomizer -c config.json index.html -n  \#atomic --bump-mq-important
```

Thoughts? Concerns?

cc @thierryk @renatoi @redonkulus @snyamathi

thierryk:
[...]

PR #385, opened 2021-02-16
Adding support for media query specificity bumps

src-code:
Implements #384.

Introduces a new option to Atomizer, `--bump-mq`:

```
  --bump-mq                           Increases specificity of media queries a small amount
```

For `--bump-mq` I'm relying on an attribute selector, `[class]`, which we know will always match (since the Atomic classname itself is applied via this attribute.)  This raises the specificity by `0,0,1,0`.

cc @snyamathi


thierryk:
👍    for:

```
--bump-mq                           Increases specificity of media queries a small amount
```

👎    for:
```
--bump-mq-important                 Override specificity of media queries by applying !important
```

I think we should offer the minimum bump we can without introducing a new "specificity context". Our use of `!important` is justified with descendant selectors but I don't see what justifies its use here. 
*What problem does `--bump-mq-important` solve that `--bump-mq` does not?*

src-code:
Sure, I can remove the support for `!important` for now, and if we determine that there's a valid use case, we can bring it back later.

redonkulus (approved):

Issue #391, opened 2021-07-08
Atomizer getConfig mutates provided object, should be pure


adamdharrington:
### Problem

We are using Atomizer in a Webpack plugin with multiple entry points feeding in a common config object for breakpoints and custom definitions. The common config object passed as second argument to `Atomizer.getConfig()` is injected with a classes property which seems to grow with additional classes each time it is called leading to later calls including all classes from previous calls.

My workaround will be to explicitly create a new object for each call however I don't believe the input object should be mutated in this case.

### Expectation

`getConfig` should be a pure function (based on it's name and implied usage)

### Reproduction

Call getConfig twice with different sources but the **same** config definition (second argument) and see the second will include all references from the first file processed.

### Fix

Return a new object and add a unit test to ensure the arguments are the same before and after the function is called.

https://github.com/acss-io/atomizer/blob/0f0fbf4d12a2b332de0d37e7dd35d9fb0663ad88/src/atomizer.js#L125-L130

PR #392, opened 2021-07-08
fix(391): don't mutate existing config in getConfig

snyamathi:
Fixes: https://github.com/acss-io/atomizer/issues/391

This PR prevents the mutation of the config which is passed into getConfig.

The documentation shows the declaration of `finalConfig` as a separate variable from `defaultConfig` so I don't believe this mutation is an intentional part of the API, just an accidental oversight. However, this could potentially cause issues for someone who was (incorrectly / unknowingly) depending on this side effect.

```js
// Generate Atomizer configuration from an array of Atomic classnames
var finalConfig = atomizer.getConfig(foundClasses, defaultConfig);
```
> https://github.com/acss-io/atomizer#api

### DIFF
--- a/src/atomizer.js
+++ b/src/atomizer.js
@@ -122,8 +122,8 @@ Atomizer.prototype.findClassNames = function (src/*:string*/)/*:string[]*/ {
  *
  * getConfig(['Op(1)', 'D(n):h']);
  */
-Atomizer.prototype.getConfig = function (classNames/*:string[]*/, config/*:AtomizerConfig*/)/*:AtomizerConfig*/ {
-    config = config || { classNames: [] };
+Atomizer.prototype.getConfig = function (classNames/*:string[]*/, existingConfig/*:AtomizerConfig*/)/*:AtomizerConfig*/ {
+    const config = existingConfig && _.cloneDeep(existingConfig) || { classNames: [] };
     // merge classnames with config
     config.classNames = this.sortCSS(_.union(classNames || [], config.classNames));
     return config;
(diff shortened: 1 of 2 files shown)


redonkulus (approved):

lingyan (approved):

src/atomizer.js
```diff
@@ -122,8 +122,8 @@ Atomizer.prototype.findClassNames = function (src/*:string*/)/*:string[]*/ {
  *
  * getConfig(['Op(1)', 'D(n):h']);
  */
-Atomizer.prototype.getConfig = function (classNames/*:string[]*/, config/*:AtomizerConfig*/)/*:AtomizerConfig*/ {
-    config = config || { classNames: [] };
+Atomizer.prototype.getConfig = function (classNames/*:string[]*/, existingConfig/*:AtomizerConfig*/)/*:AtomizerConfig*/ {
+    const config = existingConfig && _.cloneDeep(existingConfig) || { classNames: [] };
```
adamdharrington, line 126: I was in two minds about this approach, I thought we could avoid a deep clone but actually this is nice because no one consuming this can't mistakenly mutate a reference in the original config via the new one either.

Seems like the right way to go to me 👍

adamdharrington (approved):

src-code (approved):

roderickhsiao:
Haven't dig into deep details but it seems like https://github.com/acss-io/webpack-atomizer-loader/blob/master/lib/atomicLoader.ts is kind of doing side effect? tried to upgrade the latest version and the css is not generated correctly

Issue #351, opened 2019-07-31
Support for CSS variables (CSS custom properties)


chinchang:
I was trying to use this class: `Bgc(var(--clr-bg))`
And atomizer didn’t compile it.

**Expected output**:

```
.Bgc\(var\(--clr-bg\)\) {
 background-color: var(--clr-bg);
}
```

**Actual Output**:

`Nothing compiles`

CC @roderickhsiao  @src-code @renatoi

roderickhsiao:
I think this might worth to revisit 

Currently we can add a custom variable in the config like

```
clr-bg: var(--clr-bg)
```

And in Atomic we use as custom variable `Bgc(clr-bg)`.

However, we will need to manually create a CSS file for the variable somewhere.

```css
--clr-bg: #fff; 
```

thierryk:
What about going this route:

class:

```
Bgc(--clr-bg)
```

output:

```
.Bgc\(--clr-bg\) {
    background-color: var(--clr-bg);
}
```

As a side note, the variable _does not need_ to live in a CSS file, it could be declared on any node via a `style` attribute (`style="--clr-bg: teal;"`)  or inside a `<style>` block:

```
<style>
:root {
    --clr-bg: teal;
}
</style>
```

That way authors can use variables without having to edit the config nor create a stylesheet. 

Note that with custom properties **scope** is an issue, so people would need to know what they are doing.

idmytro:
I thought about this option
`Bgc$(clr-bg)` --> `background-color: var(--clr-bg)`
but this is not optimal for multiple arguments, where for example 1 is a variable, but 2 is not.

idmytro:
Windicss has this feature and use `$` sign

https://windicss.org/guide/features.html#variables

thierryk:
FWIW, I think I prefer `Bgc(--clr-bg)` over `Bgc$(clr-bg)`.
I believe the former facilitates mixing custom properties and standard values (between the parenthesis), for example:

```
Skew(20deg,--customAngle)
```

redonkulus:
@chinchang @thierryk fixed by https://github.com/acss-io/atomizer/pull/395

PR #395, opened 2021-09-21
feat: support css variables

snyamathi:
@redonkulus @renatoi @src-code This PR (attempts to) add support for [CSS Variables](https://developer.mozilla.org/en-US/docs/Web/CSS/Using_CSS_custom_properties).

This is merely basic support, echoing the token if it starts with a double dash. There is no requirement that the variables be defined in the config since it's likely that the variables defined by some shared CSS outside of atomizer.

There are other features that I _wanted_ to accomplish, but was not able to. The only marginally important one is fallback, and even that can be done with postcss using [postcss-custom-properties](https://github.com/postcss/postcss-custom-properties).

**I'm also not sure how many folks still cater to IE11 and if we want to spend the time creating features in this repo to support IE11.**

## Fallback

Ideally we could (optionally) provide fallbacks for browsers that cannot understand css variables

```css
.C\(--brand-color\) {
  color: #400090;
  color: var(--brand-color);
}
```

The main impediment to this, however, is that all the way down to [writeBlockToCSS](https://github.com/acss-io/atomizer/blob/master/src/lib/jss.js#L167), there is an assumption that there is only one value for a given property.

Passing fallback values around becomes a bit cumbersome - though hopefully someone else can take a stab at this.

## Definitions

It would be nice to (optionally) create the variable definitions created by Atomizer itself.

```css
:root {
  --brand-color: #400090;
}
```

## Re-Define Media Queries

If Atomizer does create the definitions, it may be nice to have the definitions change depending on media queries such as

```css
@media (prefers-color-scheme: dark) {
  :root {
    --background-color: #000;
    --text-color: #fff;
  }
}
```

## Custom Definition Selectors

In the above definitions, we may (optionally) want `:root` to be some other selector.

### DIFF
--- a/src/lib/grammar.js
+++ b/src/lib/grammar.js
@@ -80,6 +80,7 @@ const GRAMMAR = {
     'IMPORTANT'     : '!',
     // https://regex101.com/r/mM2vT9/8
     'NAMED'         : '([\\w$]+(?:(?:-(?!\\-))?\\w*)*)',
+    'CSS_VARIABLE'  : '(--(?:(?:-(?!\\-))?\\w*)*)',
     'BREAKPOINT'    : '--(?<breakPoint>[a-zA-Z0-9]+)',
     'PSEUDO_CLASS'  : `(?:${  flatten(PSEUDO_CLASSES).join('|')  })(?![a-z])`,
     'PSEUDO_ELEMENT': `(?:${  flatten(PSEUDO_ELEMENTS).join('|')  })(?![a-z])`,
@@ -140,6 +141,10 @@ const VALUE_SYNTAXE = XRegExp([
     '(?<unit>',
         GRAMMAR.UNIT,
     ')?',
+    '|',
+    '(?<cssVariable>',
+        GRAMMAR.CSS_VARIABLE,
+    ')',
     '|',
     '(?<named>',
         GRAMMAR.NAMED,
--- a/src/lib/grammar.js
+++ b/src/lib/grammar.js
@@ -80,7 +80,7 @@ const GRAMMAR = {
     'IMPORTANT'     : '!',
     // https://regex101.com/r/mM2vT9/8
     'NAMED'         : '([\\w$]+(?:(?:-(?!\\-))?\\w*)*)',
-    'CSS_VARIABLE'  : '(--(?:(?:-(?!\\-))?\\w*)*)',
+    'CSS_VARIABLE'  : '(--[\\w-]+)',
     'BREAKPOINT'    : '--(?<breakPoint>[a-zA-Z0-9]+)',
     'PSEUDO_CLASS'  : `(?:${  flatten(PSEUDO_CLASSES).join('|')  })(?![a-z])`,
     'PSEUDO_ELEMENT': `(?:${  flatten(PSEUDO_ELEMENTS).join('|')  })(?![a-z])`,
--- a/tests/atomizer.js
+++ b/tests/atomizer.js
@@ -1120,7 +1120,7 @@ describe('Atomizer()', function () {
             const result = atomizer.getCss(config);
             expect(result).to.equal(expected);
         });
-        it ('has do not match bad input', function () {
+        it ('should not match bad input', function () {
             const atomizer = new Atomizer();
             const config = {
                 classNames: [
--- a/tests/atomizer.js
+++ b/tests/atomizer.js
@@ -1074,13 +1074,26 @@ describe('Atomizer()', function () {
         it ('has basic suppport for css variables', function () {
             const atomizer = new Atomizer();
             const config = {
+                breakPoints: {
+                    sm: '@media screen and (min-width:700px)'
+                },
                 classNames: [
-                    'C(--brand-color)'
+                    'C(--brand-color)',
+                    'Fz(--font-size)',
+                    'Fz(--small-font-size)--sm'
                 ]
             };
             const expected = [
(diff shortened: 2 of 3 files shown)


src/lib/grammar.js
```diff
@@ -80,6 +80,7 @@ const GRAMMAR = {
     'IMPORTANT'     : '!',
     // https://regex101.com/r/mM2vT9/8
     'NAMED'         : '([\\w$]+(?:(?:-(?!\\-))?\\w*)*)',
+    'CSS_VARIABLE'  : '(--(?:(?:-(?!\\-))?\\w*)*)',
```
snyamathi, line 83: Really not sure about this regex here, so please correct this.

snyamathi:
Fixes https://github.com/acss-io/atomizer/issues/351

redonkulus (approved):
Nice!

tests/atomizer.js
```diff
@@ -1071,6 +1071,34 @@ describe('Atomizer()', function () {
...
             expect(result).to.equal(expected);
         });
+        it ('has basic suppport for css variables', function () {
+            const atomizer = new Atomizer();
+            const config = {
+                breakPoints: {
+                    sm: '@media screen and (min-width:700px)'
+                },
+                classNames: [
+                    'C(--brand-color)',
```
src-code, line 1081: Could also add test cases for identifiers with no hyphens, or other valid chars

src-code: https://drafts.csswg.org/css-values-4/#identifier-value

src/lib/grammar.js
src-code: It seems to match anything after `--`, even spaces, so I think something is not quite right.  Also can remove the outermost capture group since we don't need to capture any portion of the match I believe

src-code: https://regex101.com/r/9K9JUE/1

tests/atomizer.js
```diff
@@ -1071,6 +1071,34 @@ describe('Atomizer()', function () {
...
+                    'Fz(--font-size)',
```
src-code, line 1082: Could also add a negative test case, eg `C(--)` should not be valid

src-code (changes requested):
Holding for some negative assertions, and also verification that the regex is matching correctly

tests/atomizer.js
snyamathi: added

tests/atomizer.js
snyamathi: Added more tests in https://github.com/acss-io/atomizer/pull/395/commits/5dc7cbc932bc7aebbcecef421bf95ea453a911eb

src/lib/grammar.js
snyamathi: I've updated the regex to just `'(--[\\w-]+)'` in https://github.com/acss-io/atomizer/pull/395/commits/5dc7cbc932bc7aebbcecef421bf95ea453a911eb

redonkulus (approved):

src/lib/grammar.js
```diff
@@ -80,6 +80,7 @@ const GRAMMAR = {
     'IMPORTANT'     : '!',
     // https://regex101.com/r/mM2vT9/8
     'NAMED'         : '([\\w$]+(?:(?:-(?!\\-))?\\w*)*)',
+    'CSS_VARIABLE'  : '(--[\\w-]+)',
```
src-code, line 83: Can't use `\w` in a character match `[]`

src-code:
```suggestion
    'CSS_VARIABLE'  : '(--[a-zA-Z0-9_-]+)',
```

tests/atomizer.js
```diff
@@ -1072,6 +1072,70 @@ describe('Atomizer()', function () {
...
+            expect(result).to.equal(expected);
+        });
+        it ('should not match bad input', function () {
+            const atomizer = new Atomizer();
+            const config = {
+                classNames: [
+                    'C(--)',
+                    'C(-- foo)',
+                    'C(--$)',
+                    'C(--)',
```
src-code, line 1130: Duplicate of line 1127

src/lib/grammar.js
src-code: I assume `_` is valid for custom properties but maybe worth double-checking?

src-code: Clarified: `\w` is valid in char selector, and is only escaped here because the regex is defined as a string to be used in a `new Regex()` constructor

src-code (approved):

tests/atomizer.js
snyamathi, line 1130:
```suggestion
```

src-code (approved):

src-code:
This is a good start.  I don't think fallback is critical, but if someone wanted to contribute support under the `--ie` flag I suppose they could.  Having Atomzier manage the variable values might be of limited benefit since it forces that configuration into Atomizer-proprietary config without much added benefit?  But if someone wants to make the case for adding it, let's discuss in GH issues.

roderickhsiao:
awesome!

redonkulus:
Our 🎁 to you @roderickhsiao ❤️

roderickhsiao:
Love you

roderickhsiao:
Not sure how valuable to support fallback value like `color: var(--my-var, inherit)`?

PR #401, opened 2022-02-14
Removing `--ie` option

fidian:
This is a suggestion for what it would take to remove `--ie` as an option, so IE < 8 hacks won't be employed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,7 +41,6 @@ Usage: atomizer [options] [path]
     -w, --watch=[target]                rebuilds when changes are detected in the file, directory, or glob (argument may be passed multiple times and is parsed for Atomic CSS classes)
 
     --rtl                               swaps `start` and `end` keyword replacements with `right` and `left`.
-    --ie                                adds old IE hacks to the output.
     --exclude=[pattern]                 pattern to exclude file to be scanned
     --bump-mq                           increases specificity of media queries a small amount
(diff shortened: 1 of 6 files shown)


thierryk:
This PR should also consider removing/editing references to "oldIE" and its fixes within https://acss.io/guides/helper-classes.html and https://acss.io/tutorials/grid-system.html.

fidian:
That's a separate repository and out of scope for this pull request. I can find no references to "oldIE" in this codebase.

thierryk:
Right, I meant we can't do this without removing all "oldIE" related stuff from [acss.io](https://acss.io) first.

Also, instead of *removing*:

```
--ie                                adds old IE hacks to the output.
```

I'd replace it with:

```
--ie                                deprecated.
```

redonkulus:
I think it would just be another PR that removes the references. The amount of people relying on this old parameter is probably very small, I don't think we should keep it in the code base to say its deprecated. Furthermore, we would probably need to bump the major version of the package to avoid breaking any people that may actually be relying on this parameter still.

redonkulus:
We are not going to merge this yet until we decide to officially remove IE support. The amount of code is small and will not be outputted unless the `--ie` option is added.

redonkulus:
FYI --ie support was removed in atomizer@3.16.0

PR #400, opened 2022-02-14
Updating packages to latest versions

fidian:
XRegExp now captures named groups under "groups"

Updating packages to eliminate vulnerable packages

PR #404, opened 2022-02-16
Package update, try 2

fidian:
A follow-up to https://github.com/acss-io/atomizer/pull/400

Updating packages.
* Chalk is not at 5.x.x branch because it's now using ESM - picked 4.x.x
* Commander's parsed options are now retrieved via a method
* XRegExp now captures named groups under "groups"
* Updating packages to eliminate vulnerable packages

### DIFF
--- a/bin/atomizer
+++ b/bin/atomizer
@@ -132,7 +132,7 @@ function buildAtomicCss(additionalFiles, done) {
 
     // Generate config from parsed src files
 
-    var filesToParse = additionalFiles.concat(programOpts.args || []);
+    var filesToParse = additionalFiles.concat(program.args || []);
     if (filesToParse.length) {
         classnames = parseFiles(filesToParse, !!programOpts.recursive);
     }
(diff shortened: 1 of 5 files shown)


src-code:
Thanks @fidian!  Will review shortly

src-code:
There still seems to be something not quite right, if I run current Atomizer against a simple test file
```html
<div class="Bgc(foobar)">
  <div class="Trfo(t,c)"></div>
</div>
```

I get this:

```
$ ./bin/atomizer test.txt 
Parsing file test.txt for Atomic CSS classes

.Trfo\(t\,c\) {
  transform-origin: top center;
}
```

But if I run it against this change, I get no output from Atomizer
```
$ ./bin/atomizer test.txt 


```

fidian:
Weirdly, it works for me.
image

fidian:
I tried it and it worked, then I deleted node_modules and package-lock.json, reinstalled, and the command no longer worked. Very bizarre since it should not have worked when I ran it earlier, but I could't scroll back far enough to figure out if I was delusional or I had really seen it working.

Added a bugfix to make it work for real and updated packages again.

src-code (approved):

PR #408, opened 2022-06-01
feat: support aspect ratio css property

redonkulus:
MDN: https://developer.mozilla.org/en-US/docs/Web/CSS/aspect-ratio

NOTE: For `1 / 1` support a custom value will need to be used.


snyamathi (approved):

src-code (approved):

roderickhsiao:
🔥

PR #410, opened 2022-06-23
feat: add more global values

redonkulus:
We currently already support `inherit`, this PR adds the remaining global values (`initial`, `revert`, `revert-layer`, `unset`).

I cleaned up the rules that were using these already.

### DIFF
--- a/src/atomizer.js
+++ b/src/atomizer.js
@@ -319,10 +319,11 @@ Atomizer.prototype.parseConfig = function (config/*:AtomizerConfig*/, options/*:
                         else {
                             warnings.push(propAndValue);
                         }
-                        value = utils.getCustomValue(config, name);
 
-                        // use global values if no custom value was found
-                        if (!value) {
+                        if (name) {
+                            value = utils.getCustomValue(config, name);
+                        } else {
+                            // use global values if no custom value was found
                             switch (matchVal.groups.named) {
                                 case 'inh':
                                     value = 'inherit';
--- a/src/atomizer.js
+++ b/src/atomizer.js
@@ -14,6 +14,14 @@ const XRegExp = require('xregexp');
 
 const RULES = require('./rules.js').concat(require('./helpers.js'));
 
+const GLOBAL_VALUES = {
+    inh: 'inherit',
+    ini: 'initial',
+    rv: 'revert',
+    rvl: 'revert-layer',
+    un: 'unset'
+};
+
 /**
  * constructor
  */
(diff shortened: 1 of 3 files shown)


src/rules.js
```diff
@@ -200,13 +205,7 @@ module.exports = [
         'allowParamToValue': false,
         'styles': {
             'aspect-ratio': '$0'
-        },
-        'arguments': [{
-            'ini': 'initial',
-            'r': 'revert',
-            'rl': 'revert-layer',
-            'u': 'unset'
```
redonkulus, line 208: This doesn't match `un` but should be fine since we just added this rule

src/rules.js
```diff
@@ -894,11 +893,7 @@ module.exports = [
             'mb': 'margin-box',
             'n': 'none',
             'pb': 'padding-box',
-            'ini': 'initial',
-            'r': 'revert',
```
redonkulus, line 898: same with these since they were just added

src/rules.js
```diff
@@ -200,13 +205,7 @@ module.exports = [
         'allowParamToValue': false,
         'styles': {
             'aspect-ratio': '$0'
-        },
-        'arguments': [{
-            'ini': 'initial',
-            'r': 'revert',
```
src-code, line 206: So this is a breaking change?  Could we just use `r` for `revert` in the new globals, or are there reasons that wouldn't be safe?

src/atomizer.js
```diff
@@ -295,10 +295,22 @@ Atomizer.prototype.parseConfig = function (config/*:AtomizerConfig*/, options/*:
                     value = `var(${matchVal.groups.cssVariable})`;
                 }
                 if (matchVal.groups.named) {
-                    // first check if 'inh' is the value
+                    // add global values first
                     if (matchVal.groups.named === 'inh') {
                         value = 'inherit';
                     }
+                    else if (matchVal.groups.named === 'ini') {
```
src-code, line 302: I guess the one risk with this is that someone is using a custom value called `ini` or `un` or etc, and that would be a breaking change

src/rules.js
redonkulus: this is a new rule we recently added so the risk is small

src/atomizer.js
```diff
@@ -295,37 +293,54 @@ Atomizer.prototype.parseConfig = function (config/*:AtomizerConfig*/, options/*:
...
+                                // we have custom but we could not find the named class name there
+                                warnings.push(propAndValue);
+                            }
                         }
-                        // we have custom but we could not find the named class name there
-                        else {
+                        value = utils.getCustomValue(config, name);
+
+                        if (!value) {
                             warnings.push(propAndValue);
```
src-code, line 323: I don't think we want to warn if they're going to end up using a global value.  Would be confusing, since they weren't trying to use a custom value in the first place

src-code: So we should only warn if `value` is still null after the following `switch`

src-code (dismissed):

src/atomizer.js
```diff
@@ -295,37 +293,52 @@ Atomizer.prototype.parseConfig = function (config/*:AtomizerConfig*/, options/*:
...
+                            }
                         }
-                        // as value
-                        else if (Object.prototype.hasOwnProperty.call(config.custom, matchVal.groups.named)) {
-                            name = matchVal.groups.named;
+                        value = utils.getCustomValue(config, name);
+
+                        if (!value) {
+                            // use global values if no custom value was found
+                            switch (matchVal.groups.named) {
```
src-code, line 320: This might be more clear as a map instead of a switch

redonkulus: a map?

src/rules.js
jitendravyas: is this not a breaking change?

redonkulus: It was a brand new rule that we just added so I was confident that changing it would not affect anyone.

PR #411, opened 2022-06-23
feat(gap): add gap and row-gap styles

redonkulus:
Support [gap](https://developer.mozilla.org/en-US/docs/Web/CSS/gap) and [row-gap](https://developer.mozilla.org/en-US/docs/Web/CSS/row-gap) CSS.


src-code (approved):

src/rules.js
```diff
@@ -1705,6 +1705,29 @@ module.exports = [
     },
     /**
     ==================================================================
+    GAP
```
nspector, line 1708: Line 957 has column gap. Should that be moved into this section?

PR #419, opened 2022-06-30
feat: add missing pseudo elements

redonkulus:
Add missing `pseudo-elements` from the MDN page. I ignored experimental elements until they are  generally available.

@src-code There are two pseudo elements that I'm not sure would work with the current grammar rules:

- `::part()` - https://developer.mozilla.org/en-US/docs/Web/CSS/::part
- `::slotted()` - https://developer.mozilla.org/en-US/docs/Web/CSS/::slotted

Both of these have parentheses and I'm not sure if atomizer supports this syntax at the moment.

### DIFF
--- a/src/lib/grammar.js
+++ b/src/lib/grammar.js
@@ -43,11 +43,16 @@ const PSEUDO_CLASSES = {
 };
 
 const PSEUDO_ELEMENTS = {
-    '::before':         '::b',
-    '::after':          '::a',
-    '::first-letter':   '::fl',
-    '::first-line':     '::fli',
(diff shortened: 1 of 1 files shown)


snyamathi (approved):

redonkulus:
We can add `::part()` and `::slotted()` if there is a request later.

thierryk:
Not sure how to make that work because "parameters" here are `<ident>` and [`<compound selector>`](https://www.w3.org/TR/selectors-4/#syntax) where we can find **spaces** and I don't see how to deal with that...

PR #422, opened 2022-07-13
feat(atomizer): Add blend modes

carriemorrison:
# Add missing blend modes:

> https://developer.mozilla.org/en-US/docs/Web/CSS/background-blend-mode
> https://developer.mozilla.org/en-US/docs/Web/CSS/mix-blend-mode
>https://drafts.fxtf.org/compositing/#mix-blend-mode

We only support 1 `background-blend-mode` since atomic only supports 1 bg image

### How Has This Been Tested?
Ran atomizer as per the testing docs:
```
➜  atomizer git:(blendModes) ./node_modules/.bin/atomizer examples/html/sample-2.html
Parsing file examples/html/sample-2.html for Atomic CSS classes

.Bgbm\(n\) {
  background-blend-mode: normal;
}
.foo .foo_Bgbm\(h\) {
  background-blend-mode: hue;
}
.bar .bar_Mbm\(s\) {
  mix-blend-mode: saturation;
}
.Mbm\(pl\) {
  mix-blend-mode: plus-lighter;
}
```
Also added a basic unit test.

### Types of changes

### Checklist

---


packages/atomizer/src/blendmodes.js
```diff
@@ -0,0 +1,18 @@
+module.exports = {
```
carriemorrison, line 1: I wasn't sure about the naming convention here since we have some dupe rules starting with same letter eg `lighten` & `luminosity` also the order is from the mdn page, I could alphasort...

packages/atomizer/tests/atomizer.js
```diff
@@ -1302,4 +1302,33 @@ describe('Atomizer()', function () {
             ]);
         });
     });
+    describe('Blend modes', function () {
```
redonkulus, line 1305: I think if we are going to start adding tests for rules, we should create a new file called `rules.js` and start adding them there. Then it can be a 1 to 1 relationship with the existing `rules.js` file.

carriemorrison: yeah, I guess we need to decide if its beneficial as @src-code had mentioned we aren't really testing anything new here, just the values added...

redonkulus: We could probably automate the whole thing but looping through the entire rules.js file and running the test for each rule and creating a snapshot or something. I think all of this could be in a separate PR since we can do it for all rules at one time.

packages/atomizer/src/blendmodes.js
redonkulus: You can follow the convention here https://acss.io/guides/syntax.html#%3Cvalue%3E

carriemorrison: > If two values share the same initial letter, then the next value in alphabetical order is [abbreviated](http://en.wikipedia.org/wiki/Abbreviation) in [contracted](http://en.wikipedia.org/wiki/Contraction_%28grammar%29) form.

That's not something I am aware of, so would screen be `sc` or more like `scrn`?

redonkulus: Since these are values, I think `sc` is short and succinct.

redonkulus (approved):

PR #445, opened 2022-07-17
Add outline-offset rule

a-gunderin:
Add outline-offset rule
https://developer.mozilla.org/en-US/docs/Web/CSS/outline-offset

Outline-offset is a rule that is quite often used to stylize inputs and other elements.

I've tested it locally on my pet project.


redonkulus (approved):
We are in the process of adding a bunch of new rules. Thanks for helping!

PR #448, opened 2022-07-18
refactor: remove --ie feature support

### LINKED ISSUE #220: Document what the --ie is and isn't for, or remove it
Right now the `--ie` flag fixes a few small things, but doesn't do a whole lot.  It's unclear in the docs what one should expect `--ie` to do, and what it should leave to the user to solve via post-processors.  (I seem to recall there were specific reasons we chose to introduce `--ie` rather than leave it completely up to post-processors, but the reasons now escape me.  Maybe it was because @thierryk felt the fixes were so few and simple, it didn't make sense to require users to integrate a post-processor for them?)

It'd be good to document this, or if we find it's unnecessary to continue supporting `--ie`, we should remove it in favor of documentation that points to a post-processor like PostCSS.

@renatoi @dnewcome @thierryk

redonkulus:

---

Fixes #220
Inspired by https://github.com/acss-io/atomizer/pull/401 from @fidian 

### Description

Given that the `--ie` option was meant for IE < 8 and their support has long since been EOL'd and not widely used. It should be safe to remove the IE hacks that were previously added.

### How Has This Been Tested?

- Updated tests
- Ran atomizer with/without the flag to ensure the same output
- Ran and tested the website locally

### Types of changes

- [ ] New feature (non-breaking change which adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to change)

### Checklist


docs/assets/atomic.css
```diff
@@ -40,7 +40,7 @@
 .Bgc\(\#000\.6\) {
   background-color: rgba(0,0,0,.6);
 }
-.Bgc\(\#0280ae\), .Bgc\(brandColor\) {
+.Bgc\(\#0280ae\), .Bgc\(brandColor\), .Bgc\(\#0280ae\)\:h:hover {
```
redonkulus, line 43: @src-code I'm not sure why this collapsed into one rule

packages/atomizer/bin/atomizer
```diff
@@ -42,7 +42,7 @@ program
   .option('--exclude [pattern]', 'excluded file pattern', collect, [])
   .option('--rtl', 'swaps `start` and `end` keyword replacements with `right` and `left`')
   .option('--bump-mq', 'increases specificity of media queries a small amount')
-  .option('--ie', 'adds old IE hacks to the output')
+  .option('--ie', '[deprecated] no longer used')
```
redonkulus, line 45: Removing this option completely, breaks the command, so leaving it for now

packages/atomizer/src/helpers.js
```diff
@@ -283,9 +275,7 @@ module.exports = [
             },
             'a[class*=LineClamp]': {
                 'display': 'inline-block',
-                'display ': '-webkit-box',
-                '*display': 'inline',
-                'zoom': 1
+                'display ': '-webkit-box'
```
redonkulus, line 278: @thierryk @carriemorrison `-webkit-box` is a non standard thing, should we remove all references now?

thierryk: > `-webkit-box` is a non standard thing...

Is it?  /s

[6.1.1 Legacy Compatibility](https://www.w3.org/TR/css-overflow-3/#webkit-line-clamp)

> UAs that support [line-clamp](https://www.w3.org/TR/css-overflow-3/#propdef-line-clamp) must also support the [-webkit-line-clamp](https://www.w3.org/TR/css-overflow-3/#propdef--webkit-line-clamp)

They made a mess when they introduced vendor prefixes. i.e., MS Edge supports `line-clamp` with `-webkit-` prefix, but not with `-ms`...

I'd leave things the way they are though as this original `-webkit-` way to do line-clamp was really a hack. 
And also because for this to work, we are using `-webkit-box-orient` which is another prefixed property.

Better would be to revisit the styling/declarations in that "helper" class to use the new stuff now that we have [`line-clamp`](https://www.w3.org/TR/css-overflow-3/#propdef-line-clamp) and other goodies...

src-code (approved):

PR #456, opened 2022-07-21
fix: add style block for Cf helper

redonkulus:

---

### Description

When I [changed the `Cf` helper class](https://github.com/acss-io/atomizer/pull/448/files#diff-65751c4f29f4e5ea220240417e678b297b0c3f4e664a991370fb811fb0482349L167) to remove the IE hacks, I removed the `style` block which apparently is required to add the `:after` and `:before` styles. Without this the `Cf` class was not added in the atomic output.

### Motivation and Context

Missing `Cf` class breaks layouts.

### How Has This Been Tested?

Tested on the example site by adding `Cf` class. Also tested that adding the generated `Cf` styles to a production site fixed the layout issue.

### Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)

### DIFF
--- a/examples/css/atomic.css
+++ b/examples/css/atomic.css
@@ -4,12 +4,6 @@
 .Bgbm\(di\) {
   background-blend-mode: difference;
 }
-.Bgbm\(n\) {
-  background-blend-mode: normal;
-}
-.foo .foo_Bgbm\(h\) {
-  background-blend-mode: hue;
-}
 .Bgc\(\#000\) {
   background-color: #000;
 }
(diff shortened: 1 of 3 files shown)


examples/css/atomic.css
```diff
@@ -258,6 +243,13 @@
 .W\(9\/12\) {
   width: 75%;
 }
+.Cf:before, .Cf:after {
```
redonkulus, line 246: Classes correctly added back

PR #458, opened 2022-07-21
fix: sort arguments by keys

redonkulus:

---

### Description

Currently the Reference page outputs each class argument in the order they appear in rules.js file. As we have added new rules that share arguments, the merged object is no longer sorted. Added a `sortKeys` function to sort an objects keys after the spread operation.

Before:

Screen Shot 2022-07-21 at 12 01 49 PM

After:

Screen Shot 2022-07-21 at 12 02 07 PM

### How Has This Been Tested?

Local dev server

### Types of changes

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -92,7 +92,12 @@ module.exports = function (grunt) {
 
         watch: {
             docs: {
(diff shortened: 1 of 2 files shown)


carriemorrison (approved):
LGTM, do you need to bump the package

redonkulus:
I don't think so, since this change is really just for the website.

PR #462, opened 2022-07-27
docs navigation reorganization, add new docs

redonkulus:

---

### Description

Re-organizing the navigation to be more atomizer user friendly and less about teaching what Atomic CSS is.

### Motivation and Context

Currently, the focus on the website is teaching about Atomic CSS. This made sense in 2015 when it was a relatively new concept. However, with popular libraries like Tailwind and Tachyons, Atomic CSS is a known CSS architecture now. These docs are still really important, but not the focal point of the site, it should be about Atomizer and how to use it properly.

To help, I have re-organized the left navigation into better groupings (see screenshot):

- The current "Atomizer" guide has been renamed to Installation and includes the contents of the current Extensions guide (which now just redirects to Installation)
- Added Editor Setup and Browser Support docs for better clarification
- The new "In Depth" section includes the previous Thinking in Atomic, Class Syntax, Atomizer classes (renamed from ACSS Classes), Helper Classes and Shorthand.
- A new "Customization" section for new the Configuration doc (TBD) and Breakpoints (previously named Responsive Web Design)
- Lastly the "Tutorials" section remains unchanged although could be renamed later depending on the docs we plan to write.


docs/guides/acss-classes.md
```diff
@@ -1,169 +1,4 @@
 ---
-section: docs
-layout: docs
-title: Atomizer classes
+redirect: https://acss.io/guides/atomizer-classes.html
+layout: redirect
```
carriemorrison, line 3: nice

carriemorrison:
LGTM, this makes sense considering its 2022

carriemorrison (dismissed):

PR #471, opened 2022-08-10
dev: create search index before server start

redonkulus:

---

Make sure search index is built before dev starts.


carriemorrison (approved):
NICE 👯

PR #472, opened 2022-08-10
feat: add content to config

redonkulus:

---

### Description

Adds a new `content` property to the Atomizer configuration file. This is an optional property that when set, allows the user to define all their project path files in the atomizer config file. No longer needed to pass in the files via the cli argument.

Other changes introduced in this PR:

- Atomizer will automatically look for a `atomizer.config.js` at the root of the project. If the `--config path/to/file` is passed in, then that will take precedence just like before.
- The `--watch` option was enhanced to automatically watch the files passed in or defined in the `content` property. The previous logic of being able to pass in the files to watch still works and will only add on to the list of files watched.

More details on how it works can be read in the new Content configuration readme file.

### Motivation and Context

When running atomizer for the doc site and examples directory, it was slightly annoying to have to list out all the file patterns in the cli argument and also list them out in the `--watch` command when watching for changes. Defining them in the configuration file keeps the configuration co-located with the other project settings. Also auto watching the files found with just `--watch` was a nice improvement on dev ex as well.

### How Has This Been Tested?

My previous PR (#470) added a baseline of tests for all current cli options. This PR just modifies the `--watch` command to watch for changes from the files found. The previous functionality still remains.

### Types of changes

### Checklist


docs/configuration.md
```diff
@@ -76,6 +80,19 @@ module.exports = {
 
 The `H(uh)` will get `height: 79px` and `C(primary)` will get color: `#f6a1e1`.
 
+### Classnames
```
redonkulus, line 83: Moved this down because its not used as often

packages/atomizer/bin/atomizer
```diff
@@ -106,35 +102,10 @@ function parseFile (file, recursive, dir) {
 }
 
 function buildAtomicCss(additionalFiles, done) {
-    // Custom rulesets
```
redonkulus, line 109: This was moved outside of this function. This function is called anytime a new file is added or removed from the watcher. Since the list of rules doesn't change based on the files parsed, it didnt't make sense to add this each time. It will get run once when the script is executed.

packages/atomizer/bin/atomizer
```diff
@@ -189,6 +160,56 @@ function triggerBuild(state) {
...
+    if (!fs.existsSync(configFile)) {
+        console.error('Configuration file ' + chalk.cyan(configFile) + ' not found.');
+        return;
+    }
+    config = require(path.resolve(configFile));
+} else if (fs.existsSync(defaultConfigPath)) {
+    config = require(defaultConfigPath);
+}
+
+// Print help menu if no config is loaded and no arguments are passed in
```
redonkulus, line 180: Keep the logic from before to show the help screen if no config file is found and no files are passed in

packages/atomizer/bin/atomizer
```diff
@@ -189,6 +160,56 @@ function triggerBuild(state) {
...
+if (empty(config) && process.argv.length <= 2) {
+    program.outputHelp();
+    return;
+}
+
+// Collect files to parse from cli args
+var filesToParse = program.args || [];
```
redonkulus, line 187: As done previously, this will collect the files passed in via the cli

packages/atomizer/bin/atomizer
```diff
@@ -189,6 +160,56 @@ function triggerBuild(state) {
...
+
+// Add more src files from "content" property of config
+if (config.content && Array.isArray(config.content)) {
+    filesToParse = config.content
```
redonkulus, line 191: The special sauce to glob the patterns and concat to the filesToParse var

packages/atomizer/bin/atomizer
```diff
@@ -204,14 +225,23 @@ if (typeof programOpts.helpersNamespace !== 'undefined') {
     options.helpersNamespace = programOpts.helpersNamespace;
 }
 
-if (programOpts.watch && programOpts.watch.length) {
+// Setup Watcher
+if (programOpts.watch === true || programOpts.watch.length) {
```
redonkulus, line 229: The main change for the `--watch` option which will now automatically watch for the files parsed. But still allows explicit watching as before.

snyamathi (approved):

carriemorrison (approved):

PR #451, opened 2022-07-19
feat(atomizer): updating the box alignment logic

carriemorrison:
I confirm that this contribution is made under a BSD license and that I have the authority necessary to make this contribution on behalf of its copyright owner.

---

### Description

*W3C has been developing the Box Alignment Module, which establishes a common set of alignment properties and values for use across multiple box models, including flex, grid, table and block.*
I like where they are going with grouping alignments so makes sense (to me) to follow their syntax.

e.g.

*Value:* `normal | <baseline-position> | <content-distribution> | <overflow-position>? <content-position>`

# [Box alignment model](https://www.w3.org/TR/css-align-3/)

[ ]  *baseline position*: `first baseline` & `last baseline` not ready for use
https://www.w3.org/TR/2021/WD-css-align-3-20211224/#typedef-baseline-position
https://caniuse.com/mdn-css_properties_align-items_flex_context_last_baseline

[x] *Content Distribution*
https://www.w3.org/TR/2021/WD-css-align-3-20211224/#typedef-content-distribution

[ ] *overflow-position*: not ready for use
https://www.w3.org/TR/2021/WD-css-align-3-20211224/#typedef-overflow-position
https://caniuse.com/mdn-css_properties_align-items_flex_context_safe_unsafe

[ ] *content-position*: `start` and `end` not ready for use
https://www.w3.org/TR/2021/WD-css-align-3-20211224/#typedef-content-position
https://caniuse.com/mdn-css_properties_align-items_flex_context_start_end

[x] *self-position*: content-position + `self-end` & `self-start`
https://www.w3.org/TR/2021/WD-css-align-3-20211224/#typedef-self-position

[ ] Legacy values for justify-items - they may get dropped..

## Added helpers to match the updated specs eg 

```
const { baselinePosition, contentDistribution, contentPosition } = require('./boxAlignment');
const selfPosition = {...contentPosition, 'se': 'self-end', 'ss': 'self-start'};
```
**note:** I commented out some of the unusable values in these helpers - we could simply remove them for now too...

### Output

```
baselinePosition { 
    'b': 'baseline' 
}
contentDistribution {
    'sa': 'space-around',
    'sb': 'space-between',
    'se': 'space-evenly',
    'st': 'stretch'
}
contentPosition { 
    'c': 'center',
    'fe': 'flex-end',
    'fs': 'flex-start'
}
selfPosition {
    'c': 'center',
    'fe': 'flex-end',
    'fs': 'flex-start',
    'se': 'self-end',
[...]

PR #477, opened 2022-08-16
fix(atomizer): match 's' usage for justify-content

carriemorrison:

---
Hmm this is a fun one, we can't use `contentPosition.s = 'start` since it seems a long time back 's' was used for stretch instead of `st` like all the other properties.

https://github.com/acss-io/atomizer/pull/451/files#diff-b791d16fb36949f46d3a573d06f9dfb355562781b8668369f35b746841cbf444R1551

Since this was [just added](https://github.com/acss-io/atomizer/commit/db2db1302e77426bc2698267edcfeac6b8296261#diff-73f5f0f4c1d3e6a68a8412e0f9203506d0d0652530c80e88e596681a1e4ea784R18), we could change the value to `sta`, even though I dislike it...

I'm not sure of a way we can be backwards compat and have consistency...

### DIFF
--- a/docs/_layouts/docs.html
+++ b/docs/_layouts/docs.html
@@ -2,7 +2,7 @@
 layout: default
 ---
 
-<div id="toggleMenuWrapper" class="docs-page D(f) Ai(s) Gp(40px) Mt(40px) Mb(50px) Mx(a) Px(10px) Maw(1280px)--sm">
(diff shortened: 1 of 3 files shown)


carriemorrison:
Screen Shot 2022-08-16 at 11 27 25 AM

redonkulus:
Given its very low usage, I'm ok being consistent and changing the value to match the other properties. I checked internally and we do not use this property.

carriemorrison:
OK, So I went ahead and remove the old backwards compat rule from the code....

carriemorrison:
Screen Shot 2022-08-16 at 12 56 25 PM

redonkulus (approved):

PR #507, opened 2022-09-15
feat(docs): add copy button to code blocks

redonkulus:

---

Using JS to add a simple copy button to the code blocks to make it easier for users to grab the code. Took inspiration from Github's copy code feature.

copycopy

### DIFF
--- a/docs/guides/syntax.md
+++ b/docs/guides/syntax.md
@@ -198,13 +198,21 @@ A suffix mapped to a [pseudo-element](https://developer.mozilla.org/en-US/docs/W
 -   `ph` for `::placeholder`
 -   `s` for `::selection`
 
+This example uses the `content` property to add open and closed quotes with the `::b` and `::a` pseudo-elements.
+
+```html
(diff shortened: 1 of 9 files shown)


docs/guides/syntax.md
```diff
@@ -198,13 +198,21 @@ A suffix mapped to a [pseudo-element](https://developer.mozilla.org/en-US/docs/W
 -   `ph` for `::placeholder`
 -   `s` for `::selection`
 
+This example uses the `content` property to add open and closed quotes with the `::b` and `::a` pseudo-elements.
+
+```html
+<q class="Cnt(oq)::b Cnt(cq)::a">Hello world!</q>
```
carriemorrison, line 204: maybe do an example without `<q>` since that adds quotes :) use  `<blockquote>` perhaps

redonkulus: Ya blockquote has default styles in our stylesheet, I can just use simple `<p>` tag

PR #509, opened 2022-09-16
docs: add dark mode document

redonkulus:

---

Add dark mode guide for users.

### DIFF
--- a/docs/guides/dark-mode.md
+++ b/docs/guides/dark-mode.md
@@ -1,0 +1,158 @@
+---
+description: How to use Atomizer to enable dark mode on your site.
+layout: docs
+section: docs
+title: Dark Mode
+---
+
+Dark mode has become a standard feature across many websites. OS level support and the [`prefers-color-scheme`] CSS media feature has made it even easier to enable dark mode.
+
+Atomizer does not provide out of the box support for dark mode because it is not required. You can combine the `prefers-color-scheme` CSS preference and the existing [`context`] class support to add dark mode to your website.
+
+Continue reading to learn how to use each concept.
+
+## Browser Basics
+
+All modern browsers [support](https://caniuse.com/prefers-color-scheme) the [`prefers-color-scheme`] CSS media feature. The feature works by defining or overwriting an existing CSS property via the `@media (prefers-color-scheme: dark) {}` directive.
+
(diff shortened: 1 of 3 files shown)


docs/guides/dark-mode.md
```diff
@@ -0,0 +1,158 @@
...
+
+Atomizer does not provide out-of-the-box support for dark mode because it is not required. You can combine the `prefers-color-scheme` CSS preference and the existing [`context`] class support to add dark mode to your website.
+
+Continue reading to learn how to use each concept.
+
+## Browser Basics
+
+All modern browsers [support](https://caniuse.com/prefers-color-scheme) the [`prefers-color-scheme`] CSS media feature. The feature works by defining or overwriting an existing CSS property via the `@media (prefers-color-scheme: dark) {}` directive.
+
+In the example below, we set an initial style for a `div` with the `.intro` class. We overwrite the style when the OS or user-agent preference is set to `dark`.
```
carriemorrison, line 18: maybe we should have this last since it requires `JS` we could show the `prefers-color-scheme` logic first since that is pure css....

docs/guides/dark-mode.md
```diff
@@ -0,0 +1,168 @@
...
+
+When the `dark` class is added earlier in the HTML tree, the styles will change to `background-color: #333;` and `color: #fff`.
+
+```html
+<body class="dark">
+    <div class="Bgc(#eee) C(#000) dark_Bgc(#333) dark_C(#fff)">Hello World!</div>
+</body>
+```
+
+<p class="noteBox info">By default, without the <code>dark</code> class on the page, the dark mode classes will not be applied (even if the user has the dark mode scheme enabled). You would need to add the <code>dark</code> class to your website to "enable" the dark mode classes. Read the <a href="#enabling-dark-mode">Enabling Dark Mode</a> section for tips on how to add the class to your project.</p>
```
carriemorrison, line 110: we could also add an example that does the var changes in css eg

```
/* overrides for dark mode */
.dark {
    --background-color: #333;
    --text-color: #fff;
}
```

carriemorrison (dismissed):
LGTM :)

docs/guides/dark-mode.md
redonkulus: I was thinking about that but it's annoying when the override styles have to be duplicated in the style sheet.

PR #527, opened 2022-10-28
feat: support :nth-child odd/even pseudo-class

redonkulus:
Requested via https://github.com/orgs/acss-io/discussions/525

Adds pseudo-class support for `:nth-child(odd)` and `:nth-child(even)`. Note: Functional syntax support was not added for now, as the additional `()` maybe cause parser problems.

Screenshot of how it would be used:
Screen Shot 2022-10-28 at 8 24 25 AM


docs/guides/syntax.md
```diff
@@ -151,13 +151,43 @@ Optional.
...
+-   `:placeholder-shown` => `:ps`,
+-   `:read-only` => `:ro`,
+-   `:read-write` => `:rw`,
+-   `:required` => `:req`,
+-   `:right` => `:r`,
+-   `:root` => `:rt`,
+-   `:scope` => `:s`,
+-   `:target` => `:t`,
+-   `:valid` => `:va`,
+-   `:visited` => `:vi`,
```
snyamathi, line 190: Would this be easier digested as a table

snyamathi (dismissed):

docs/guides/syntax.md
redonkulus: Hmm perhaps, I can try moving it to that

redonkulus: fixed https://github.com/acss-io/atomizer/pull/527/commits/3c5b556fe169d4f946dc290ccdee13964ea82300

snyamathi (approved):

carriemorrison:
Do we think having only `odd` or `even` is useful? very limited use cases.. like styling table rows lol

redonkulus:
The :nth-child fractional support would require atomizer parser updates to support the paranthesis in the class name. We haven't done that before for any other pseudo class or element.

carriemorrison:
I guess these type of atomic rules are only useful for v simple loops since your going to add the SAME rule to each child element, like in your example.

```
 <ul>
    <li class="Bgc(red):odd">First item</li>
    <li class="Bgc(red):odd">Second item</li>
    <li class="Bgc(red):odd">Third item</li>
</ul>
```

Otherwise you could just alter the markup

```
            <ul>
                <li class="Bgc(red)">First item</li>
                <li class="">Second item</li>
                <li class="Bgc(red)">Third item</li>
            </ul>
```

I guess you could output both without any logic whatsoever

```
<ul>
    <li class="Bgc(red):odd Bgc(blue):even">First item</li>
    <li class="Bgc(red):odd Bgc(blue):even">Second item</li>
    <li class="Bgc(red):odd Bgc(blue):even">Third item</li>
</ul>
```

I personally would just use css for advanced logic, if we feel  having `odd` and `even` only is helpful then sure we can add this :)

redonkulus:
Ya I agree odd/even are not really that beneficial when outputting on every child. I think having support for `:nth-child()` would be better. However, that will take some time to update the syntax and test is thoroughly. I'm going to close this PR for now.

PR #532, opened 2022-11-16
feat: add strict option

snyamathi:

---

### Description

Add an option `strict` which will bail (exit code 1) if there is missing context which is required to safely generate the CSS

### Motivation and Context

A missing custom class config will log a warning and not generate any CSS. Not a huge deal.

A missing breakpoint will not log a warning, but it _will_ generate CSS which is missing the media query. This can be extremely dangerous since the CSS will silently be applied globally as opposed to the intended media query, however accidentally omitted.

While this currently only applies to missing break points, I'd like to reserve the option to expand this to cover other missing context without a breaking change.

### How Has This Been Tested?

It's behind a config and tested with one of our apps.

### Types of changes

### Checklist

### DIFF
--- a/packages/atomizer/package.json
+++ b/packages/atomizer/package.json
@@ -24,8 +24,8 @@
         "*.d.ts"
     ],
     "scripts": {
-        "lint": "eslint --ext .js .",
-        "lint:fix": "eslint --ext .js . --fix",
+        "lint": "eslint --ext .js . && prettier . --check",
+        "lint:fix": "eslint --ext .js . --fix && prettier . --write",
         "test": "jest tests/"
     },
     "bin": {
--- a/packages/atomizer/src/atomizer.js
+++ b/packages/atomizer/src/atomizer.js
@@ -28,7 +28,9 @@ const GLOBAL_VALUES = {
  * @param {Array<import('atomizer').AtomizerRule>} [rules] List of custom rules
  */
 function Atomizer(options, rules) {
-    this.verbose = (options && options.verbose) || false;
+    options = options || {};
+    this.strict = options.strict || false;
+    this.verbose = options.verbose || false;
     this.rules = [];
     // we have two different objects to avoid name collision
     this.rulesMap = {};
@@ -481,6 +483,8 @@ Atomizer.prototype.parseConfig = function (config, options) {
  */
 Atomizer.prototype.getCss = function (config, options) {
     const jss = {};
+    const isStrict = this.strict;
+    const isVerbose = this.verbose;
     let content = '';
     let breakPoints;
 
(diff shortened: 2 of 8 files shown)


packages/atomizer/src/atomizer.js
```diff
@@ -532,6 +536,20 @@ Atomizer.prototype.getCss = function (config, options) {
                 }
 
                 const breakPoint = breakPoints && breakPoints[treeo.breakPoint];
+                if (treeo.breakPoint && !breakPoint) {
+                    const message = [
+                        `Class \`${treeo.className}\` contains breakpoint \`${treeo.breakPoint}\` which does not exist and must be manually added to your config file:`,
```
carriemorrison, line 541: nice

carriemorrison (dismissed):
Tks @snyamathi this is very helpful.

packages/atomizer/src/atomizer.js
```diff
@@ -532,6 +536,20 @@ Atomizer.prototype.getCss = function (config, options) {
...
+                        '"breakPoints": {',
+                        `    "${treeo.breakPoint}": <YOUR-CUSTOM-VALUE>`,
+                        '}',
+                    ].join('\n');
+                    if (isStrict) {
+                        console.error('Error:', message);
+                        process.exit(1);
+                    } else if (isVerbose) {
+                        console.warn('Warning:', message);
```
snyamathi, line 550: Note that this will still write the "bad" CSS. Without a valid breakpoint, `W(100%)--sm'` will yield

```css
.W\(100\%\)--sm {
  width: 100%;
}
```

Again the issue being that there is no media query so this is applied in all instances rather than

```css
@media(...) {
  .W\(100\%\)--sm {
    width: 100%;
  }
}
```

packages/atomizer/package.json
```diff
@@ -24,8 +24,8 @@
         "*.d.ts"
     ],
     "scripts": {
-        "lint": "eslint --ext .js .",
-        "lint:fix": "eslint --ext .js . --fix",
+        "lint": "eslint --ext .js . && prettier . --check",
+        "lint:fix": "eslint --ext .js . --fix & prettier . --write",
```
redonkulus, line 28: Only one `&`?

docs/guides/syntax.md
```diff
@@ -81,43 +81,43 @@ Optional.
...
-| `:read-only`         | `:ro`   |
-| `:read-write`        | `:rw`   |
-| `:required`          | `:req`  |
-| `:right`             | `:r`    |
-| `:root`              | `:rt`   |
-| `:scope`             | `:s`    |
-| `:target`            | `:t`    |
-| `:valid`             | `:va`   |
-| `:visited`           | `:vi`   |
+| Pseudo-Element       | Suffix |
```
redonkulus, line 84: Rebase from `main` will pull in this change for you

redonkulus (commented):
Changes look good to me, just two minor comments.

packages/atomizer/package.json
snyamathi, line 28:
```suggestion
        "lint:fix": "eslint --ext .js . --fix && prettier . --write",
```

typo

docs/guides/syntax.md
snyamathi: done

redonkulus (approved):

PR #536, opened 2022-11-23
feat: remove loader-utils from webpack-atomizer-loader

redonkulus:
getOptions was removed from loader-utils@3.x. Its part of the loader API now, so no longer need this package. Bumping major version to be safe.

Replaces https://github.com/acss-io/atomizer/pull/530


tom76kimo:
🚀

PR #539, opened 2022-11-28
feat: support outline-color, outline-style, outline-width

redonkulus:

---

Adds support for the missing outline properties (color, style, width). Also did some refactoring to re-use the border values across rules.

### DIFF
--- a/packages/atomizer/src/border.js
+++ b/packages/atomizer/src/border.js
@@ -1,0 +1,19 @@
+module.exports.borderStyles = {
+    a: 'auto',
+    d: 'dotted',
+    da: 'dashed',
+    do: 'double',
+    g: 'groove',
+    h: 'hidden',
+    i: 'inset',
+    n: 'none',
+    o: 'outset',
+    r: 'ridge',
+    s: 'solid',
+};
+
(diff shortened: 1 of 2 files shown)


packages/atomizer/src/border.js
```diff
@@ -0,0 +1,19 @@
+module.exports.borderStyles = {
```
carriemorrison, line 1: nice

packages/atomizer/src/rules.js
```diff
@@ -2494,6 +2387,16 @@ module.exports = [
...
+        matcher: 'Oc',
+        allowParamToValue: true,
+        styles: {
+            'outline-color': '$0',
+        },
+        arguments: [colors],
+    },
     {
         type: 'pattern',
         name: 'Outline-offset',
```
carriemorrison, line 2402: interesting that this existed without the other rules

carriemorrison (approved):

packages/atomizer/src/rules.js
redonkulus: Ya it must have been an outside contribution

PR #542, opened 2022-11-29
feat: support touch-action and will-change

redonkulus:

---

Added support for:
- touch-action -> Tcha()
- will-change -> Wc(s). This one can also accept custom values based on the CSS property used (not to mention double values too). These use cases are best resolved via custom atomizer config variables.

### DIFF
--- a/packages/atomizer/src/rules.js
+++ b/packages/atomizer/src/rules.js
@@ -3946,7 +3946,7 @@ module.exports = [
         type: 'pattern',
         name: 'Will change',
         matcher: 'Wc',
-        allowParamToValue: false,
+        allowParamToValue: true,
         styles: {
             'will-change': '$0',
         },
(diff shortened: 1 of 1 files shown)


packages/atomizer/src/rules.js
```diff
@@ -3911,6 +3939,27 @@ module.exports = [
...
+    */
+    {
+        type: 'pattern',
+        name: 'Will change',
+        matcher: 'Wc',
+        allowParamToValue: false,
+        styles: {
+            'will-change': '$0',
+        },
+        arguments: [
```
carriemorrison, line 3953: will change has many args, do we need `allowParamToValue` ?

https://developer.mozilla.org/en-US/docs/Web/CSS/will-change

carriemorrison: I guess custom would work here?

carriemorrison (approved):

packages/atomizer/src/rules.js
redonkulus: Changed it to `true` to make it easier to use. Custom var would still be needed for double values

PR #547, opened 2022-11-30
test: add snapshots for all rules

redonkulus:

---

As I've been adding/modifying rules, there really wasn't a way to verify the syntax is correct. I added a simple jest test that will iterate over the rules and dynamically build each atomizer class name based on the available arguments.

NOTE: One caveat is that for rules that support multiple values (e.g. `place-content: baseline center`), I only test the first property of the first argument with each property of the second argument. Testing each combination of both would be overkill. Also, the one downside of this approach that potentially some combinations of values may conflict with each other if actually used on a page.

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -7,7 +7,9 @@ node_modules/
 vendor/
 
 # files
+.*
 package-lock.json
+*.snap
 
 # docs
 app/assets/css/atomic.css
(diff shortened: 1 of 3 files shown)


carriemorrison (dismissed):
This is super helpful! 👯

PR #552, opened 2022-12-02
feat: add scrollbar-* properties

redonkulus:

---

Requested by @nspector. 

Adding `scrollbar-*` properties

### DIFF
--- a/packages/atomizer/src/rules.js
+++ b/packages/atomizer/src/rules.js
@@ -3300,6 +3300,74 @@ module.exports = [
     },
     /**
     ==================================================================
+    SCROLLBAR-COLOR
+    ==================================================================
+    */
+    {
+        type: 'pattern',
+        name: 'Scrollbar color',
+        matcher: 'Sbc',
+        allowParamToValue: false,
+        styles: {
+            'scrollbar-color': '$0 $1',
+        },
+        arguments: [
+            {
+                a: 'auto',
+                ...colors,
+            },
+            {
+                a: 'auto',
+                ...colors,
+            },
+        ],
+    },
+    /**
+    ==================================================================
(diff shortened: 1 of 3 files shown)


packages/atomizer/src/rules.js
```diff
@@ -3300,6 +3300,74 @@ module.exports = [
...
+        styles: {
+            'scrollbar-gutter': '$0 $1',
+        },
+        arguments: [
+            {
+                a: 'auto',
+                s: 'stable',
+            },
+            {
+                be: 'both-edges',
```
redonkulus, line 3344: @carriemorrison I'm not sure if this `both-edges` value applies to both `auto` and `stable` or just `stable`. If its the latter, I could just modify this rule to support a `sbe` value which maps to "stable both-edges".

The spec for reference: https://www.w3.org/TR/2021/WD-css-overflow-3-20211223/#propdef-scrollbar-gutter

carriemorrison:
Screen Shot 2022-12-02 at 4 18 51 PM

carriemorrison:
Same for https://developer.mozilla.org/en-US/docs/Web/CSS/scrollbar-color

carriemorrison:
Maybe leave in draft form till fully adopted?

nspector (approved):

redonkulus:
Good point about the draft status of `scrollbar-gutter` and `scrollbar-color`. Generally, we only add properties once the reach recommendation status. It looks like there is a ticket open to add to webkit but it hasn't gained traction: https://bugs.webkit.org/show_bug.cgi?id=167335

redonkulus:
Given these rules are not recommended standards yet, let's hold on including them. At least until Safai supports them too.

PR #553, opened 2022-12-05
fix(atomizer-plugins): run atomizer on file updates

redonkulus:

---

Fixes the atomizer-unplugin library to run atomizer on application file updates. Code copied from @maheshbansod [investigation](https://github.com/maheshbansod/atomizer-vite/blob/main/atomizerPlugin.ts).

### DIFF
--- a/.changeset/silent-needles-flash.md
+++ b/.changeset/silent-needles-flash.md
@@ -1,0 +1,5 @@
+---
+"atomizer-plugins": patch
+---
+
+fix(atomizer-plugins): run atomizer on file updates
--- a/docs/integrations/vite.md
+++ b/docs/integrations/vite.md
@@ -49,7 +49,7 @@ import atomizerConfig from './atomizer.config.js';
 
 const atomizerPlugin = vite({
     config: atomizerConfig,
-    outfile: 'atomizer.css',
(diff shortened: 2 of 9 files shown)


maheshbansod:
@redonkulus in my project, I'm planning to make this so that `writeFile` in `transform` will invoke during dev mode only - without the dev mode flag, transform will exit early and `buildEnd` will do the writeFile.
I think this will make build faster, since it doesn't have to update the file on every transform.

redonkulus:
@maheshbansod how will you check dev mode without the flag, `NODE_ENV=development`?

Will you keep the `transformIndexHtml` hook as well as `buildEnd`? I think if you move it off of `transform` then the `debounce` may not be needed after all.

redonkulus:
@maheshbansod I've added the dev flag check using `NODE_ENV` since vite already uses that standard. So now it will only call `writeFile()` in dev mode and leave writing to file in production for `buildEnd`.

redonkulus:
Ok please test this out locally. I will merge and release tomorrow. Thanks!

maheshbansod:
@redonkulus tested the code on my side in a Vite plugin. working well.

PR #554, opened 2022-12-05
feat: add text-decoration-* rules

redonkulus:

---

This PR adds the remaining CSS rules to minimally match what Tailwind supports today. There are hundreds more that can be added but at least aligns up better with other contemporaries.

- text-decoration-*
- break-*
- backdrop-filter: *

### DIFF
--- a/packages/atomizer/src/rules.js
+++ b/packages/atomizer/src/rules.js
@@ -34,6 +34,7 @@
  **/
 const { baselinePosition, contentDistribution, contentPosition } = require('./boxAlignment');
 const blendModes = require('./blendmodes');
+const breakValues = require('./break');
 const mixBlendModes = Object.assign(blendModes, { pd: 'plus-darker', pl: 'plus-lighter' });
 const colors = require('./colors');
 const { borderStyles, borderWidths } = require('./border');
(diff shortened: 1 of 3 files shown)


carriemorrison:
just wondering about `Bf` for break-before... since we have `Ba` for break-after

redonkulus:
> just wondering about `Bf` for break-before... since we have `Ba` for break-after

Oops, that was a typo, should've been `Bb`. Fixed now.

carriemorrison (dismissed):
LGTM

packages/atomizer/src/rules.js
```diff
@@ -667,6 +667,127 @@ module.exports = [
...
+        allowParamToValue: true,
+        styles: {
+            'backdrop-filter': 'contrast($0)',
+        },
+    },
+    // dropshadow
+    {
+        type: 'pattern',
+        name: 'Backdrop Drop shadow (filter)',
+        matcher: 'BkdpDropshadow',
```
carriemorrison, line 723: these are looking v long, should we follow 

`Bxsh(<custom-param>) box-shadow:`

though the filters are also long so I guess we match them...

carriemorrison (approved):
LGTM

PR #620, opened 2023-03-31
feat(containers): Added container queries

carriemorrison:

---

### Description

Added support for container queries, using the same breakpoint logic since that is vague in description and allows users to config the required string. It is possible to use NON named containers though I don't think that is very useful for atomic css. Using named containers relies on adding extra config as described in the updated documentation.

### Motivation and Context

Because Container queries! but Seriously very helpful to be able to use them.

### How Has This Been Tested?

Added an example file using named containers.

### Types of changes

### Checklist

*reference*

Screenshot 2023-03-31 at 2 40 18 PM

*Example gif*

container

*Documentation*
0 0 0 0_4000_breakpoints html


docs/breakpoints.md
```diff
@@ -97,3 +97,84 @@ When using explicit breakpoints, use multiple classes to have styles applied in
...
+We have added support for container queries using the same syntax as media queries, since you are able to define what the rule is, just ensure the names are different to the media query rules.
+
+```js
+breakPoints: {
+    sm: '@media screen and (min-width: 700px)',
+    mw300: '@container (max-width: 300px)',
+},
+```
+
+Same logic as above - append `--<breakpoint name>` to any Atomic class to associate that styling with the breakpoint of your choice. For example, `Fz(1rem)--mw300` will create the following rule in the related container query:
```
redonkulus, line 113:
```suggestion
Same logic as above - append `--<breakpoint name>` to any Atomizer class to associate that styling with the breakpoint of your choice. For example, `Fz(1rem)--mw300` will create the following rule in the related container query:
```

docs/breakpoints.md
```diff
@@ -97,3 +97,84 @@ When using explicit breakpoints, use multiple classes to have styles applied in
...
+
+```css
+@container width300 (max-width: 300px) {
+  .Fz(1rem)--contmw300 {
+    font-size: 1rem;
+  }
+}
+```
+
+```html
```
redonkulus, line 155: Would be good to add a sentence above just to differentiate this example from the one above.

.changeset/tall-lions-give.md
```diff
@@ -0,0 +1,5 @@
+---
+"atomizer": minor
+---
+
+Added container queries
```
redonkulus, line 5:
```suggestion
feat: Added container queries support
```

packages/atomizer/src/rules.js
```diff
@@ -1343,6 +1343,38 @@ module.exports = [
     },
     /**
     ==================================================================
+    CONTAINERS
+    ==================================================================
+    // TODO: Discuss `n` , `nor`, `z`
```
redonkulus, line 1348: you can remove this, what you have is good

packages/atomizer/src/rules.js
```diff
@@ -1343,6 +1343,38 @@ module.exports = [
...
+        name: 'Container Type',
+        matcher: 'ContType',
+        allowParamToValue: false,
+        styles: {
+            'container-type': '$0',
+        },
+        arguments: [
+            {
+                is: 'inline-size',
+                n: 'none',
```
carriemorrison, line 1370: this follows existing pattern
```
                n: 'none',
                nor: 'normal',
```

packages/atomizer/src/rules.js
```diff
@@ -1343,6 +1343,38 @@ module.exports = [
...
+                nor: 'normal',
+                s: 'size',
```
carriemorrison, line 1372: we only have 1 other 'size' and that exists in same object as a `style`
```
            {
                n: 'none',
                st: 'strict',
                c: 'content',
                z: 'size',
                l: 'layout',
                s: 'style',
                p: 'paint',
            },
```

we could follow it and use `z` or keep `s` as we also have `is` for `inline-style`

redonkulus (approved):

PR #912, opened 2024-10-02
Add word-wrap: anywhere support

nspector:

---

### Description

Adds support for `word-wrap: anywhere` - https://developer.mozilla.org/en-US/docs/Web/CSS/overflow-wrap

### Motivation and Context

Adds support for a supported value of word-wrap

### How Has This Been Tested?

Snapshots have been updated.

### Types of changes

### Checklist

### DIFF
--- a/packages/atomizer/src/rules.js
+++ b/packages/atomizer/src/rules.js
@@ -4430,7 +4430,7 @@ module.exports = [
             {
                 bw: 'break-word',
                 n: 'normal',
-                a: 'anywhere'
+                a: 'anywhere',
             },
         ],
     },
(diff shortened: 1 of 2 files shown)


packages/atomizer/src/rules.js
```diff
@@ -4430,6 +4430,7 @@ module.exports = [
             {
                 bw: 'break-word',
                 n: 'normal',
+                a: 'anywhere',
```
snyamathi, line 4433: > To prevent overflow, an otherwise unbreakable string of characters — like a long word or URL — may be broken at any point if there are no otherwise-acceptable break points in the line. No hyphenation character is inserted at the break point. Soft wrap opportunities introduced by the word break are considered when calculating min-content intrinsic sizes.

snyamathi (approved):

redonkulus (approved):

PR #970, opened 2025-04-22
Fix postcss atomizer to avoid duplicate CSS

carystanley:

---

### Description

Fix `postcss-atomizer` so that you can control which css file and where the CSS is added.

### Motivation and Context

Currently the atomizer css is added to every single file processed by postcss, this can cause duplication if you have many css files in your project. This change allows you to control which file and where the generated css will go.

### How Has This Been Tested?

In an environment containing multiple css files I confirmed that the atomizer CSS is only added to css files where the `@atomizer;` rule is.

### Types of changes

-   [ ] New feature (non-breaking change which adds functionality)
-   [x] Breaking change (fix or feature that would cause existing functionality to change)

### Checklist

-   [ ] I have added tests to cover my changes.

### DIFF
--- a/.changeset/metal-flies-look.md
+++ b/.changeset/metal-flies-look.md
@@ -1,0 +1,5 @@
+---
+"postcss-atomizer": patch
+---
+
+BREAKING CHANGE: postcss atomizer to avoid duplicate CSS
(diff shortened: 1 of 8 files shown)


redonkulus (approved):
Thanks for the fix!

PR #1023, opened 2026-01-15
perf(jss): optimize combineSelectors from O(n²) to O(n) (#234)

### LINKED ISSUE #234: Performance of JSS.combineSelectors() is poor
From @3den in pr #233:

> Problem: extracted seems to be an array of objects, the challenge we have is to find the keys that map to the same value in order to create the selector and to do that we take current O(n^2).

> Suggestion: we can think about this as an inverted hash where the key is the style we are applying and the value is an array of selectors that match that value every time you find a new selector we just push it to the corresponding style, the object could look something like:

> ```js
{
   'display: inline-block;': ['.D\(ib\)', '.IbBox', 'a[class*=LineClamp]']
}
```
> ps: I don't know if that suggestion would work better so fell free to ignore it.

jitendravyas:
Replace nested loop implementation with inverted hash approach for combining CSS selectors with the same property/value pair.

Before: O(n²) - nested loops comparing all pairs
After: O(n) - single pass using hash map lookup

This significantly improves performance for large stylesheets with many atomic classes.

Fixes #234

---

### Description

_Describe your changes in detail._

### Motivation and Context

_Why is this change required? What problem does it solve?_

### How Has This Been Tested?

_Please describe in detail how you tested your changes. Include details of your testing environment, and the tests you ran to see how your change affects other areas of the code, etc._

### Types of changes

-   [ ] New feature (non-breaking change which adds functionality)

### Checklist

-   [ ] I have updated the documentation accordingly.
-   [ ] I have read the [CONTRIBUTING](https://github.com/acss-io/atomizer/blob/main/CONTRIBUTING.md) document.
-   [ ] I have added tests to cover my changes.


redonkulus:
Did you test this to make sure nothing broke? Do we have tests that cover this use case?

What speed improvement do we see from this?

jitendravyas:
no i did not test. how to do it?

